PR #14, opened 2023-11-03
Fixed a bug where solo aspect wasn't correctly parsed

tomas-davidovic:
When the `filter_data` is just a string, then the `filter_aspect` on the next line was a list and would assign `aspect` (which has to be a global of some sort I guess, or a literal) instead of the `filter_aspect`. The result was that it wouldn't pick up aspects.

In general, the two lines that array-ify the `filter_data` seems superfluous, one should be enough (if not an array, wrap in an array)?

### DIFF
--- a/src/item/filter.py
+++ b/src/item/filter.py
@@ -111,7 +111,7 @@ def should_keep(item: Item):
     if item.aspect and filter_aspects is not None:
         for filter_data in filter_aspects:
             filter_aspect = [filter_data] if isinstance(filter_data, str) else filter_data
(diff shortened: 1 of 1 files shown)


aeon0 (changes requested):
Good catch, thast a bug. Lets clean this up. Remove L113 and do:
` aspect_name, *rest = filter_data if isinstance(filter_data, list) else [filter_data]`

tomas-davidovic:
Updated as requested

aeon0 (approved):

PR #15, opened 2023-11-03
Added elementalists_aspect to the list of aspects that have variable …

tomas-davidovic:
…number on a second position

The aspect says that skills cast above 100 Mana (first number) gain XX% critical strike chance. So it is one of those "second number is the important one" cases


aeon0 (approved):
Great. There are probably a couple more that have their variable numbers as second or third. Probably should add this to the json file to easier adapt this instead of somewhere in the code. But works for now.

PR #16, opened 2023-11-03
Adjust aspect value based on the item type

tomas-davidovic:
The current version reads the aspect variable value as is.

But those values are scaled by 1.5x for Amules and by 2x for two handed weapons.
I propose to scale them back down to the "canonical" values.

(I am not running Rogue, so I don't know whether Bows and Crossbows scale to 2x as well, please adjust accordingly)

### DIFF
--- a/src/item/read_descr.py
+++ b/src/item/read_descr.py
@@ -250,6 +250,13 @@ def read_descr(rarity: ItemRarity, img_item_descr: np.ndarray) -> Item:
             idx = 0
         found_value = _find_number(concatenated_str, idx)
 
+        # Scale the aspect down to the canonical range if found on an item that scales it up
+        if item.type is ItemType.Amulet:
+            found_value /= 1.5
+        # Possibly add Bow and Crossbow if those scale it up as well
+        if item.type in [ItemType.Axe2H, ItemType.Sword2H, ItemType.Mace2H, ItemType.Scythe, ItemType.Polearm, ItemType.Staff]:
+            found_value /= 2
+
         if found_key is not None:
             item.aspect = Aspect(found_key, found_value, concatenated_str)
         else:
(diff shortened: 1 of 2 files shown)


src/item/read_descr.py
```diff
@@ -250,6 +250,13 @@ def read_descr(rarity: ItemRarity, img_item_descr: np.ndarray) -> Item:
             idx = 0
         found_value = _find_number(concatenated_str, idx)
 
+        # Scale the aspect down to the canonical range if found on an item that scales it up
+        if item.type is ItemType.Amulet:
+            found_value /= 1.5
+        # Possibly add Bow and Crossbow if those scale it up as well
+        if item.type in [ItemType.Axe2H, ItemType.Sword2H, ItemType.Mace2H, ItemType.Scythe, ItemType.Polearm, ItemType.Staff]:
```
johannes-do:
```suggestion
        if item.type in [ItemType.Bow, ItemType.Crossbow, ItemType.Axe2H, ItemType.Sword2H, ItemType.Mace2H, ItemType.Scythe, ItemType.Polearm, ItemType.Staff]:
```

aeon0 (changes requested):
Yes, should be done! Bows and Crossbows also scale by 2. I added it in a suggestion.

tomas-davidovic:
Sorry, fighting with the system a bit. I accepted the suggestion and it keeps saying Changes requested so trying to figure out what's up. :-D

aeon0 (approved):

aeon0:
You have to adapt the read_descr_test.py.  As this is an amulet: https://github.com/aeon0/d4lf/blob/main/test/assets/item/read_descr_legendary_1920x1080_1.png. It now has different value for the aspect. (correct value for the aspect that is :)

Adapt here: https://github.com/aeon0/d4lf/blob/2dbe9678ac9620d4832713927c3161bb2991575a/test/item/read_descr_test.py#L44

tomas-davidovic:
Adapted

aeon0 (approved):

PR #29, opened 2023-11-04
Dev/added all aspects with numbers in pos1 and 2

tomas-davidovic:
Does what it says on the tin, the position is based on the first variable number.


aeon0 (approved):
awesome, thank you

PR #74, opened 2023-11-15
More robust home directory handling

xenpie:


xenpie:
For me the application wasn't working initially because I changed my Windows account name when switching from a Microsoft to a local account. Although my Username was changed the directory still uses the previous name and therefore f"C:/Users/{os.getlogin()}/ didn't return the correct path. Using os.path.expanduser('~') should always return the correct home directory.

xenpie:
Alternatively you could also use os.environ['USERPROFILE'] or perhaps pathlib.Path.home() but I haven't tested those.

aeon0 (approved):
Cool, didnt know about that

PR #176, opened 2024-02-11
Moved minAffixCount to any_of instead of globally

chrisoro:
Hello, good sir. Firstly, I must commend you on your awesome project. But I was irritated with the process of specifying which items I wanted to have multiple affixes from.

Let's say I desire a Scythe with the following criteria:
- It must have intelligence.
- It should have 3 of the following: basic_skill_damage, core_skill_damage, damage_to_injured_enemies, damage_while_healthy.

The configuration would resemble this:
```
- TestOld:
  itemType: [ scythe ]
  minPower: 700
  affixPool:
  - [ intelligence ]
  - any_of:
    - [ basic_skill_damage ]
    - [ core_skill_damage ]
    - [ damage_to_injured_enemies ]
    - [ damage_while_healthy ]
  - any_of:
    - [ basic_skill_damage ]
    - [ core_skill_damage ]
    - [ damage_to_injured_enemies ]
    - [ damage_while_healthy ]
  - any_of:
    - [ basic_skill_damage ]
    - [ core_skill_damage ]
    - [ damage_to_injured_enemies ]
    - [ damage_while_healthy ]
    minAffixCount: 4
```

Coming from Path of Exile, the trading website allows you to specify your "any_of" with a count, so I wanted to implement it like this:
```
- TestNew:
  itemType: [ scythe ]
  minPower: 700
  affixPool:
  - [ intelligence ]
  - any_of:
    - [ basic_skill_damage ]
    - [ core_skill_damage ]
    - [ damage_to_injured_enemies ]
    - [ damage_while_healthy ]
    minAffixCount: 3
```

This is what my change does. If you like it, I will also add the description to the readme.

Other changes I have done:
- simplified (at least from my point of view) the logic when to mark an affix as found. If you don't like the change, I can revert
- Removed minAffixCount from global item level since it doesn't make sense anymore. You either specify the up to 4 specific affixes you want or you use (multiple) any_of with a dedicated count. The minAffixCount is derived from the number of rules you define on an item. If you don't like the change, I can also make it backward compatible since the old logic was basically any_of with minAffixCount for that condition
- I think I also fixed a bug where you 'deleted the affixes from the item' after you matched it to a rule. This is problematic if someone stupid enought did multiple any_of in a ways that only 1 or 2 of them match. If those 1 or 2 got deleted, then they remaining any_ofs wouldn't match. Might be seen as a bug or as a user error ;)

### DIFF
--- a/src/item/filter.py
+++ b/src/item/filter.py
@@ -1,16 +1,17 @@
-from item.models import Item
-from dataclasses import dataclass, field
-import yaml
-import json
 import os
 import time
+from dataclasses import dataclass, field
 from pathlib import Path
-from logger import Logger
+
+import yaml
+
 from config import Config
 from dataloader import Dataloader
+from item.data.affix import Affix
 from item.data.item_type import ItemType
 from item.data.rarity import ItemRarity
-from item.data.affix import Affix
+from item.models import Item
+from logger import Logger
 
 
 @dataclass
@@ -228,28 +229,27 @@ class Filter:
 
         for affix in filter_affix_pool:
             if isinstance(affix, dict) and "any_of" in affix:
+                filter_min_affix_count = affix["minAffixCount"] if "minAffixCount" in affix and affix["minAffixCount"] is not None else 1
                 any_of_matched = self._match_affixes(affix["any_of"], item_affix_pool)
-                if len(any_of_matched) > 0:
-                    name = any_of_matched[0]
-                    item_affix_pool = [a for a in item_affix_pool if a.type != name]
-                    matched_affixes.append(name)
+                if len(any_of_matched) >= filter_min_affix_count:
+                    matched_affixes.extend(any_of_matched)
             else:
                 name, *rest = affix if isinstance(affix, list) else [affix]
                 threshold = rest[0] if rest else None
                 condition = rest[1] if len(rest) > 1 else "larger"
 
-                item_affix_value = next((a.value for a in item_affix_pool if a.type == name), None)
-                if item_affix_value is not None:
-                    if (
-                        threshold is None
-                        or (isinstance(condition, str) and condition == "larger" and item_affix_value >= threshold)
-                        or (isinstance(condition, str) and condition == "smaller" and item_affix_value <= threshold)
-                    ):
-                        item_affix_pool = [a for a in item_affix_pool if a.type != name]
-                        matched_affixes.append(name)
-                elif any(a.type == name for a in item_affix_pool):
-                    item_affix_pool = [a for a in item_affix_pool if a.type != name]
-                    matched_affixes.append(name)
+                item_affix = next((a for a in item_affix_pool if a.type == name), None)
+                if item_affix is None:  # affix not found
+                    continue
+                # affix has no value or value isn't as desired
+                if item_affix.value is not None and not (
+                    threshold is None
+                    or (isinstance(condition, str) and condition == "larger" and item_affix.value >= threshold)
+                    or (isinstance(condition, str) and condition == "smaller" and item_affix.value <= threshold)
+                ):
+                    continue
+                # all other cases are good
+                matched_affixes.append(name)
         return matched_affixes
 
     def should_keep(self, item: Item) -> FilterResult:
@@ -284,14 +284,14 @@ class Filter:
                         filter_min_affix_count = (
                             filter_data["minAffixCount"]
                             if "minAffixCount" in filter_data and filter_data["minAffixCount"] is not None
-                            else 0
+                            else 1
                         )
                         power_ok = self._check_power(filter_data, item)
                         type_ok = self._check_item_type(filter_data, item)
                         if not power_ok or not type_ok:
                             continue
                         matched_affixes = self._match_affixes(filter_data["affixPool"], item.affixes)
-                        affixes_ok = filter_min_affix_count is None or len(matched_affixes) >= filter_min_affix_count
+                        affixes_ok = len(matched_affixes) >= filter_min_affix_count
                         inherent_ok = True
                         matched_inherent = []
                         if "inherentPool" in filter_data:


aeon0:
The idea/usecase was that you might want to have sth like this:

```yaml
  - Armor:
      itemType: [chest armor, pants]
      minPower: 725
      affixPool:
        - total_armor
        - damage_reduction_from_close_enemies
        - damage_reduction_from_distant_enemies
        - any_of:
          - fire_resistance
          - shadow_resistance
          - poison_resistance
          - lightning_resistance
          - cold_resistance
      minAffixCount: 3
```

So have any one of the resistances, but maximum 1 resistance. I couldn't test yet, but how would items behave with your implementation:
- [total_armor, damage_reduction_from_close_enemies, fire_resistance] (should match)
- [total_armor, fire_resistance, shadow_resistance] (should not match)

chrisoro:
Ok, I see your point. minAffixCount still makes sense globally. I added it back.

minAffixCount on any_of level is now just a helper to remove the need to copy paste duplicated any_ofs

```
Affixes:
  - Armor:
      itemType: [ chest armor, pants ]
      minPower: 725
      affixPool:
        - total_armor
        - damage_reduction_from_close_enemies
        - any_of:
            - fire_resistance
            - shadow_resistance
            - poison_resistance
            - lightning_resistance
            - cold_resistance
          minAffixCount: 2
      minAffixCount: 3
```

instead of
```
Affixes:
  - Armor:
      itemType: [ chest armor, pants ]
      minPower: 725
      affixPool:
        - total_armor
        - damage_reduction_from_close_enemies
        - any_of:
            - fire_resistance
            - shadow_resistance
            - poison_resistance
            - lightning_resistance
            - cold_resistance
        - any_of:
            - fire_resistance
            - shadow_resistance
            - poison_resistance
            - lightning_resistance
            - cold_resistance
      minAffixCount: 3
```

src/item/filter.py
```diff
@@ -228,28 +229,27 @@ def _match_affixes(self, filter_affix_pool: list, item_affix_pool: list[Affix])
...
         for affix in filter_affix_pool:
             if isinstance(affix, dict) and "any_of" in affix:
+                filter_min_affix_count = affix["minAffixCount"] if "minAffixCount" in affix and affix["minAffixCount"] is not None else 1
                 any_of_matched = self._match_affixes(affix["any_of"], item_affix_pool)
-                if len(any_of_matched) > 0:
-                    name = any_of_matched[0]
-                    item_affix_pool = [a for a in item_affix_pool if a.type != name]
-                    matched_affixes.append(name)
+                if len(any_of_matched) >= filter_min_affix_count:
+                    matched_affixes.extend(any_of_matched)
```
johannes-do, line 235: Should all of any_of_matched be included or rather: `any_of_matched[:filter_min_affix_count]? Usecase: Someone wants two have 2 resistances on the boots. Now they would need to copy the res block twice. With the new implementaiton the goal would be to set minAffixCount: 2 inside the any_of block.

But if we do this also boots with 3 or 4 resistances would show up as matched.

src/item/filter.py
```diff
@@ -284,14 +284,14 @@ def should_keep(self, item: Item) -> FilterResult:
                         filter_min_affix_count = (
                             filter_data["minAffixCount"]
                             if "minAffixCount" in filter_data and filter_data["minAffixCount"] is not None
-                            else 0
+                            else 1
```
johannes-do, line 287: I see the point to make it behave the same way as the minAffixCount in any_of blocks. Just note to self: Put a note in the patch notes so users know that this has changed!

src/item/filter.py
```diff
@@ -228,28 +229,27 @@ def _match_affixes(self, filter_affix_pool: list, item_affix_pool: list[Affix])
...
+                item_affix = next((a for a in item_affix_pool if a.type == name), None)
+                if item_affix is None:  # affix not found
+                    continue
+                # affix has no value or value isn't as desired
+                if item_affix.value is not None and not (
+                    threshold is None
+                    or (isinstance(condition, str) and condition == "larger" and item_affix.value >= threshold)
+                    or (isinstance(condition, str) and condition == "smaller" and item_affix.value <= threshold)
+                ):
+                    continue
```
johannes-do, line 250: thats nicer

aeon0 (changes requested):

src/item/filter.py
chrisoro: Well depends. We can do the following just fine:
```
affixPool:
    - any_of:
      - total_armor
      - damage_reduction_from_close_enemies
      minAffixCount: 2
    - any_of:
      - fire_resistance
      - shadow_resistance
      - poison_resistance
      - lightning_resistance
      - cold_resistance
      minAffixCount: 2
```
It all depends on what the user wants to do, it will not be the solution for everything but for some specific use cases  where you don't care about about 2-3 affixes and just want them out of a pool of many.

src/item/filter.py
chrisoro: Should also mention the release notes that affixes are no longer removed from pool after match. Previously something like
```
affixPool:
    - any_of:
      - fire_resistance
      - shadow_resistance
      - poison_resistance
      - lightning_resistance
      - cold_resistance
      minAffixCount: 1
    - any_of:
      - fire_resistance
      - shadow_resistance
      - poison_resistance
      - lightning_resistance
      - cold_resistance
      minAffixCount: 1
```

Would have not matched an item with just fire res, since you removed the affix from the pool on match. This is not done anymore. This is somewhat conceived since you now would just do

```
affixPool:
    - any_of:
      - fire_resistance
      - shadow_resistance
      - poison_resistance
      - lightning_resistance
      - cold_resistance
      minAffixCount: 2
```

and be done with it

src/item/filter.py
johannes-do: Imho this defeats the whole point of the any_of block. You might as well just add them outside the any_of. Someone asked about "how can I make certain affixes obligatory". This is currently possible by doing sth like this:

```yaml
BootsTest2:
  itemType: boots
  minPower: 700  
  affixPool:
    - [intelligence]
    - any_of:
      - [movement_speed]
      - [poison_resistance]
      - [cold_resistance]
      - [lightning_resistance]
  minAffixCount: 2
  ```
When we dont remove the additional matches from the any_of block. A boot with 2 light res and cold res but no intelligence will also match the filter. Meaning, there is no difference using the any_of block or not.

johannes-do: Maybe we shouldnt call the any_of `minAffixCount` but rather `maxMatches`. That way if previously you had to copy the res block 2 times cause you were looking for 2 of any resistances you can now do `maxMatches: 2`.

chrisoro: I had the same idea yesterday. I think the [PoE trade site](https://www.pathofexile.com/trade) filtering is perfect.
The problem here is that every item in affixPool is evaluated using 'and' but minAffixCount makes it an implicit 'or'.

Let me clean this up and provide a different pull request with the PoE logic.

chrisoro:
As discussed, obsolete due to: https://github.com/aeon0/d4lf/pull/179

PR #179, opened 2024-02-16
Refactor Item Filtering

chrisoro:
As discussed via Discord, I'll put this up for you to look at.


chrisoro:
As discussed, will move parts of it to smaller prs and we do the interface change at the end of the season

PR #182, opened 2024-02-16
Small lootfilter improvements

chrisoro:
- Reduced idle time after switching tab
- No longer move mouse to a random position after switching tabs
- Simplified lootfilter.check_items a bit


src/loot_filter.py
```diff
@@ -82,14 +82,12 @@ def check_items(inv: InventoryBase):
         res = Filter().should_keep(item_descr)
         matched_any_affixes = len(res.matched) > 0 and len(res.matched[0].matched_affixes) > 0
         Logger.debug(f"  Runtime (Filter): {time.time() - start_filter:.2f}s")
-        if not res.keep:
-            keyboard.send("space")
-            wait(0.13, 0.14)
-        elif res.keep and (matched_any_affixes or item_descr.rarity == ItemRarity.Unique):
+        # mark everything as junk
```
johannes-do: Items that match affixes will be marked as favorite. Other items such as matched aspect, elixier, material, etc. will be kept. Everything else will be marked as junk.

If we do this, the items that only match the aspect for example will also be junked.

aeon0 (changes requested):

src/loot_filter.py
chrisoro: Yup you are correct. Will fix

aeon0 (approved):

PR #188, opened 2024-02-19
Use pydantic for validation

chrisoro:
- Add new folder config housing several modules regarding config reading. This replaces the old config.py
  - Everything is pydantic based and verified on load. This means that the user gets feedback asap and us developers don't need to typecheck every few lines
- `config/game.ini` is no longer needed and has been moved to code
- `config/res/game*.ini` are no longer needed. The whole mechanism has been reworked. The base full hd config has been moved to code and using are calculated on runtime using `config.ui.ResManager.set_resolution`
- Changed whole code base accordingly
- Set detect_window_thread to daemon, so it's nicer to debug. The whole thread handling here should be refactored.. ;)


src/item/find_descr.py
```diff
@@ -24,30 +23,36 @@ def choose_best_result(res_left: SearchResult, res_right: SearchResult) -> Searc
...
 
-    roi_right = copy(Config().ui_roi["rel_descr_search_right"])
+    roi_right = ResManager().roi.rel_descr_search_right
     roi_right[0] += anchor[0]
-    ok, roi_right = fit_roi_to_window_size(roi_right, Config().ui_pos["window_dimensions"])
+    ok, roi_right = fit_roi_to_window_size(roi_right, ResManager().pos.window_dimensions)
     if ok:
         res_right = search(ref=refs, inp_img=img, roi=roi_right, threshold=0.8, mode="best")
 
+    # TODO why is this happening?!
```
chrisoro: @aeon0 I have triple checked my code and can't find the error. Can you help please?

Comparing this branch and main, debugging shows the same data and type. But for some reason both res are always None using this branch.

Ah, I have an idea. Might be the transform singleton. Will check it later

chrisoro: Found it

aeon0:
looks good

aeon0 (approved):

PR #189, opened 2024-02-23
Support all resolutions

chrisoro:
- Refactored scaling to support all resolutions. Resolution scaling and black bars are calculated according to D4 specifications. Everything greater than 16 / 10 aspect ratio is supported. Lower than 16/10 will introduce black bars on top/bottom (easy to calc) but some strange ui shift which I'm not able to calculate. Since these res are very uncommon anyways, I don't think this is a problem
- Recalculate scaling every time a the d4 window is found/moved/resized/...
- Only read config from the user folder if the params.ini file there is not empty

### DIFF
--- a/src/template_finder.py
+++ b/src/template_finder.py
@@ -106,7 +106,9 @@ class SearchArgs:
 def stored_templates() -> dict[Template]:
     paths = []
     templates = {}
-    for path in [f"assets\\templates_{Cam().res_p}"]:
+    # TODO check if it works for higher res
+    closest = min([1080, 1440, 1600, 2160], key=lambda x: abs(x - int(Cam().res_p[:-1])))
+    for path in [f"assets\\templates_{closest}p"]:
         paths += list_files_in_folder(path)
     for file_path in paths:
         file_name: str = os.path.basename(file_path)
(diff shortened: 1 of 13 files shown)


src/template_finder.py
```diff
@@ -106,7 +106,9 @@ def wait_for_update(img: np.ndarray, roi: list[int] = None, timeout: float = 3,
 def stored_templates() -> dict[Template]:
     paths = []
     templates = {}
-    for path in [f"assets\\templates_{Cam().res_p}"]:
+    # TODO check if it works for higher res
```
chrisoro, line 109: must still check

chrisoro:
@aeon0 for some reason some unit tests are failing. I double checked the implementation and it should be good. Can you please take a look? why some tests fail?

chrisoro:
We could also keep the highest resolution template and then just scale it to the others. This would resolve the need for templates for all resolutions, right?

src/template_finder.py
chrisoro: it does work if it is close enough. But never worse than old implementation

chrisoro:
All done

aeon0 (approved):
awesome!

PR #197, opened 2024-04-10
Add possibility to select specific stash tabs

chrisoro:
Instead of just specifying the amount, you can now specify the actual tabs.

This has been asked for many times in discord


aeon0 (approved):
Works, thx

PR #198, opened 2024-04-18
Fix check_chest_tabs and added Glacial Fissure sigil dungeon

BimmerHex:
When we set the check_chest_tabs value in the config to 1.2, the tab index started at 1 and started by checking the 2nd tab. Therefore, by setting it to -1, it worked correctly by starting from index 0

Added Glacial Fissure sigil dungeon for The Beast in the Ice Dungeon

### DIFF
--- a/assets/lang/enUS/sigils.json
+++ b/assets/lang/enUS/sigils.json
@@ -119,7 +119,8 @@
         "sealed_archives": "sealed archives",
         "seaside_descent": "seaside descent",
         "shifting_city": "shifting city",
-        "whispering_vault": "whispering vault"
+        "whispering_vault": "whispering vault",
+        "glacial_fissure": "glacial_fissure"
     },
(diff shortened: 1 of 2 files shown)


BimmerHex:
Frankly, I haven't seen any problems until this update. I made a few changes locally, so when I did this update it gave me this error. So it might be my fault. I am not sure. I'm playing at 1920x1080 resolution.

chrisoro:
I just tested again with fhd and it works. Might be some local changes on your end. Can you test unchanged main to verify?

aeon0:
I am also on 1920x1080 and the current main also works for me. The "-1" is already taken care of in the model validator here:
https://github.com/aeon0/d4lf/blob/67b5eb5fb9c67ea3ad32a3fb8243e6bc1d9092cf/src/config/models.py#L73

PR #199, opened 2024-04-20
Add Pydantic for profile validation

chrisoro:
- Implemented Pydantic for profile validation, ensuring that all user input conforms to expectations and eliminating the need for type checking or error handling after initial validation. Additionally, Pydantic provides improved output for users to assist them in defining valid profiles.
- Refactored `dataloader.py` to accommodate validation within Pydantic Models.
- Updated README documentation.
- Overhauled profiles distributed with the installation. Outdated profiles have been replaced by an example profile demonstrating correct syntax and usage, with a focus on structure rather than specific content (or a particular character/build).
- Renamed all Models in `config/modely.py` to include the suffix `Model`, clearly indicating that these are validated user inputs.
- Restructured `item/filter.py` for improved clarity and reduced complexity. The code path is now easier to understand, with fewer side effects.
  - Introduced filter types to `affixPool`, starting with the initial type called `count`. `count` functions similarly to the `count` filter in the PoE trade website.
  - This refactor facilitates the addition of other possible filter types without breaking existing ones. The previous logic, which mixed `and/or` in one type, was not future-proof and had flaws.
- Expanded unit test coverage for all new features and refactors.


aeon0 (approved):
Smooth, tested, works.

PR #200, opened 2024-04-22
Add support for all resolutions

chrisoro:
- Set the default for all templates/masks/positions to UHD (Ultra High Definition). 
- Updated template matching from hardcoded to scaled template matching. UHD resources will now be appropriately scaled and matched. 
  - Because of the scaling, I had to slightly relax some of the pattern matching thresholds.
  - Prioritize searching for the medium icons first and then the small icons. Due to scaling, they now appear quite similar, and by checking the smaller ones first, we might miss some if font is set to medium.
- Centralized internal configuration data in `config.data.py`.
- Removed all non-UHD assets.


chrisoro:
@aeon0 , please review. There are 3 failed tests, all indicating compression artifacts in the test images. Could you confirm if all test images were saved without compression?

aeon0 (approved):
works great

PR #201, opened 2024-04-26
Fix / improve Model validation

chrisoro:
- Prohibit extra sections in either YAML or INI files other than the ones specified via the Model
- Addressed a bug in `item.filter.Filter.load_files` that incorrectly assigned an empty filter to a profile


aeon0 (approved):
Approved

PR #202, opened 2024-04-26
Fix value comparisons

chrisoro:
- Adjusted all value comparisons to now include the specified number
  - For example, 4 and "larger" will include 4.
  - Similarly, 4 and "smaller" will also include 4.


aeon0 (approved):
Better safe than sorry

aeon0 (approved):

PR #204, opened 2024-05-01
Added new bullet points

chrisoro:
- Added templates for greater affixes and tempered affixes (medium).
- Note: I didn't make further changes, but we need to make more changes for the next season.
  - Specifically, pants no longer have inherents, see https://d4builds.gg/database/gear-affixes/
  - Unfortunately, I couldn't add unit tests since we are missing all new affix texts, and I only have screenshots with unknown affixes compared to live.

image


aeon0:
Yeah, I think we will have to wait until the new season drops and then just go from there.

PR #205, opened 2024-05-11
Rework Aspects for Season 4

chrisoro:
- Reworked Aspects filtering:
  - It no longer makes sense to filter for specific values. The focus is now on salvaging for materials or codex upgrades, or selling for gold.
  - Removed old filtering based on specific values.
  - Introduced new filtering based on upgrade state.
  - Configured via `keep`, which can be set to `all`, `none`, or `upgrade`.
  - Only added an icon for the medium font since there is no screenshot available for the small font.
- Updated `README.md` accordingly.
- Housekeeping:
  - `item.models.Item`: Sorted alphabetically and changed all usages to use kwargs.
  - `config.models.ProfileModel`: Added default values for `Aspects` and `Sigils` for better structure and ease of use.


README.md
```diff
@@ -80,33 +80,27 @@ filter, or even split the same type of filter over multiple files. Ultimately, a
 your `params.ini` will be used to determine if an item should be kept. If one of the profiles wants to keep the item, it
 will be kept regardless of the other profiles. Similarly, if a filter is missing in all profiles (e.g., there is no `Sigils` section in any profile), all corresponding items (in this case, sigils) will be kept.
 
-### Aspects
+### Affix / Aspects Filter Syntax
```
johannes-do, line 85: Its only Affix filtering, no more aspects, no?

aeon0 (changes requested):

README.md
chrisoro: Can be used for unique items

aeon0 (approved):
good!

PR #209, opened 2024-05-15
S4 quick fixes

aeon0:


src/config/models.py
```diff
@@ -82,6 +82,12 @@ def set_defaults(cls, data: "AffixFilterCountModel") -> "AffixFilterCountModel":
             data["maxCount"] = len(data["count"])
         return data
 
+    @field_validator("minCount")
+    def min_count_validator(cls, minCount: int) -> int:
```
chrisoro, line 86: should we also do max count?

src/item/descr/texture.py
```diff
@@ -85,6 +84,14 @@ def find_aspect_search_area(img_item_descr: np.ndarray, aspect_bullet: TemplateM
...
-def find_codex_upgrade_icon(img_item_descr: np.ndarray) -> bool:
+def find_codex_upgrade_icon(img_item_descr: np.ndarray, aspect_bullet: TemplateMatch) -> bool:
+    top_limit = img_item_descr.shape[0] // 2
+    right_limit = img_item_descr.shape[1] // 2
+    if aspect_bullet is not None:
+        top_limit = aspect_bullet.center[1]
+    cut_item_descr = img_item_descr[top_limit:, :right_limit]
     # TODO small font template fallback
-    return search(["codex_upgrade_icon_medium"], img_item_descr, 0.8, use_grayscale=True, mode="first").success
+    result = search(["codex_upgrade_icon_small"], cut_item_descr, 0.78, use_grayscale=True, mode="first")
```
chrisoro: I think we should always do medium first and then small. scaled medium works most of the time for small font but scaled small never works for medium font. I changed this for all other find_* in a previous pr

src/config/models.py
johannes-do: max count should also be at least 1 I guess then, right?

chrisoro: max count should be [1,5]. upper boundary is not as important as it simply wouldn't match. but we can catch this user error

PR #210, opened 2024-05-15
Fix keep_aspects

chrisoro:
- Added missing value loader for keep_aspects. It was always using the default


aeon0:
Missed that on the review 🤦

PR #226, opened 2024-05-18
Adding maxroll.gg scraping support

Maximo1491:
Imports a build and creates a yaml profile based on the either a build guide or planner from Maxroll

Not quite sure on the flow of how to import these profiles. I've added the hotkey early so you can import builds without diablo running, have tested a few profiles and seem to import fine, ignoring uniques for now.


README.md
```diff
@@ -224,13 +225,13 @@ effects, you only have to specify the thresholds that you want to apply.
 <details><summary>Config Examples</summary>
 
 ```yaml
-# Take all uniques with item power > 800 
+# Take all uniques with item power > 800
```
maximo1491, line 258: I'm assuming this is from the settings being applied from the readme
"Setup VS Code by using the black formater extension. Also turn on "trim trailing whitespaces" is VS Code settings."

Maximo1491:
Also do wonder if this would just be better as a seperate tool rather than trying to roll it all into the same package

aeon0:
Nice one! will test. I generally have the experience if we put it in a standalone it will not get updated properly after some time. I'd prefer to have it all in one mono repo.

aeon0:
Tested, works nicely. One thing on the logger output. Can you remove your info logger and instead add it to the table and move the table above the checking for the d4 window? Then its more visible and cleaner:

sth like this (obviously remove the table further down):
image

aeon0:
Let me know when you think the PR is ready

Maximo1491:
Moved the table creation up a bit, think that should be it for now.

src/config/models.py
```diff
@@ -109,18 +109,19 @@ class AdvancedOptionsModel(_IniBaseModel):
     exit_key: str
     log_lvl: str = "info"
     process_name: str = "Diablo IV.exe"
+    import_build: str
```
chrisoro: please keep it sorted alphabetically :)

src/config/loader.py
```diff
@@ -45,6 +45,7 @@ def _load_params(self):
             exit_key=self._select_val("advanced_options", "exit_key"),
             log_lvl=self._select_val("advanced_options", "log_lvl"),
             process_name=self._select_val("advanced_options", "process_name"),
+            import_build=self._select_val("advanced_options", "import_build"),
```
chrisoro: please keep it sorted alphabetically :)

src/config/models.py
```diff
@@ -109,18 +109,19 @@ class AdvancedOptionsModel(_IniBaseModel):
...
     run_filter: str
     run_scripts: str
     scripts: list[str]
 
     @model_validator(mode="after")
     def key_must_be_unique(self) -> "AdvancedOptionsModel":
-        keys = [self.exit_key, self.run_filter, self.run_scripts]
+        keys = [self.exit_key, self.import_build, self.run_filter, self.run_scripts]
```
chrisoro, line 125: please keep it sorted alphabetically :)

src/config/models.py
```diff
@@ -109,18 +109,19 @@ class AdvancedOptionsModel(_IniBaseModel):
...
         if len(set(keys)) != len(keys):
             raise ValueError(f"hotkeys must be unique")
         return self
 
-    @field_validator("run_scripts", "run_filter", "exit_key")
+    @field_validator("run_scripts", "import_build", "run_filter", "exit_key")
```
chrisoro: please keep it sorted alphabetically :)

src/utils/build_importer.py
```diff
@@ -0,0 +1,116 @@
...
+from yaml import dump
+
+
+def import_build():
+    Logger.info(f"Paste maxroll.gg build guide or planner build here ie https://maxroll.gg/d4/build-guides/minion-necromancer-guide")
+    url = input()
+    if "maxroll.gg/d4/planner" not in url and "maxroll.gg/d4/build-guides/" not in url:
+        Logger.info(f"Invalid url, please use a d4 planner build on MaxRoll.gg")
+        return
+    options = webdriver.ChromeOptions()
```
chrisoro: Since the API for all webdrivers is the same, can't we just let the user decide and then use the appropriate driver? So declare an enum, add a new settings that is validated against that enum and then here, create proper options/webdriver object.

Chrome, Edge, Firefox, (Chromium) are the most used right now.

src/utils/build_importer.py
```diff
@@ -0,0 +1,116 @@
...
+    options.add_argument("--headless=new")
+    driver = webdriver.Chrome(options=options)
+    driver.get(url)
+    wait = WebDriverWait(driver, 10)
+    wait.until(EC.element_to_be_clickable((By.CLASS_NAME, "ncmp__btn")))
```
chrisoro: Starting from here, lots of magic strings. For maintainability I would suggest moving them to module level as constants.

I'm referring to class names, xpaths, etc. Basically everything you use to locate the elements

src/utils/build_importer.py
```diff
@@ -0,0 +1,116 @@
...
+    Logger.info(f"Importing {item_dict["class"]} {filter_label} build")
+
+    for item in item_slots:
+        item.click()
+        driver.find_element(By.CLASS_NAME, "d4t-title").click()
+        item_type = wait.until(EC.element_to_be_clickable((By.CLASS_NAME, "d4t-category-header"))).text
+        if item_type == "Weapon" or item_type == "Equipment":
+            # if this is just "Weapon" then the planner uses different builds at different stages of the build, so we find the active weapon for this current version of the build, and grab the name of the weapon type
+            item_type = (
+                driver.find_element(By.CLASS_NAME, "d4t-ItemLibrary")
```
chrisoro: Instead of this complex nested query, you could just provide one xpath, no?

src/utils/build_importer.py
```diff
@@ -0,0 +1,116 @@
...
+            )
+        item_properties = driver.find_elements(By.CLASS_NAME, "d4t-item-mods")
+        try:
+            item_options = driver.find_element(By.CLASS_NAME, "d4t-item-options")
+        except NoSuchElementException:
+            continue
+        # skip uniques for now, need to figure out a better way to grab these
+        if item_options.text.startswith("Unique"):
+            driver.find_element(By.CLASS_NAME, "d4t-header-title").text
+            Logger.info(f"Skipping {driver.find_element(By.CLASS_NAME, "d4t-header-title").text} unique importing currently unsupported")
```
chrisoro:
```suggestion
            Logger.warning(f"Skipping {driver.find_element(By.CLASS_NAME, "d4t-header-title").text} unique importing currently unsupported")
```
I think warning is better since the user should see that he might need to change his Unique filter

src/utils/build_importer.py
```diff
@@ -0,0 +1,116 @@
...
+        # commenting out for now, but leaving here to sort in the future
+        # item_details["tempers"] = item_properties[1] if len(item_properties) > 1 else ""
+        # item_details["aspect"] = item_properties[2] if len(item_properties) > 2 else ""
+        # item_details["gems"] = item_properties[3] if len(item_properties) > 3 else ""
+        item_dict[item_type] = item_details
+
+    driver.quit()
+    filter = convert_to_filter(item_dict)
+    with open(f"{Path(f"{os.path.expanduser("~")}/.d4lf")}/profiles/{filter_label}.yaml", "w") as file:
+        file.write(dump(filter, default_flow_style=None, sort_keys=False))
```
chrisoro: everybody loves sorting. why disable it? :)

src/utils/build_importer.py
```diff
@@ -0,0 +1,116 @@
...
+
+    print(f"Created profile {Path(f"{os.path.expanduser("~")}/.d4lf")}\\profiles\\{filter_label}.yaml")
```
chrisoro: Use logger here

src/utils/build_importer.py
```diff
@@ -0,0 +1,116 @@
...
+
+
+def translate_modifiers(mods):
```
chrisoro: only used in this module so please prefix with "_"

src/utils/build_importer.py
```diff
@@ -0,0 +1,116 @@
...
+    translated_mods = []
+    mods = mods.find_elements(By.CLASS_NAME, "d4t-property")
+    for mod in mods:
+        translated_mods += [[remove_extra_underscores(re.sub(r"[0-9]|,|\.|\+|\[|\]|-|%", "", mod.text).lower().strip().replace(" ", "_"))]]
+
+    return translated_mods
+
+
+def remove_extra_underscores(str):
```
chrisoro: only used in this module so please prefix with "_"

src/utils/build_importer.py
```diff
@@ -0,0 +1,116 @@
...
+    while i < len(str):
+        if str[i] == str[i - 1] and str[i] == "_":
+            str = str[:i] + str[(i + 1) :]
+            i -= 1
+        i += 1
+
+    return str
+
+
+def convert_to_filter(items):
```
chrisoro: only used in this module so please prefix with "_"

src/utils/build_importer.py
chrisoro: I have not tried it out but off the top of my head:
`def _remove_extra_underscores(string):
    return re.sub(r'(_)\1+', r'\1', string)`
This is the same but easier, right?

src/utils/build_importer.py
chrisoro: Instead of a dict, please use our (validated) classes. Makes it easier to maintain and adapt to future changes. Talking about `config.models.ProfileModel`. You can use the package `pydantic-yaml` to generate the yaml based on that object

chrisoro (changes requested):
Great work.
Mostly style remarks but one important change regarding maintainability and one possibile addition to support more browser than only Chrome

src/utils/build_importer.py
chrisoro: `str `is bad arg name since it shadow pythons `str`

src/config/models.py
maximo1491: Completely missed, these were alphabetical. Cheers!

src/utils/build_importer.py
maximo1491: This was to keep the profiles how they're created now, i.e. itemType then the affixPool. But I do also prefer keeping it sorted, so I'll remove this :)

src/utils/build_importer.py
maximo1491: Regex commands still baffle me! Yes, this seems to work though, so I've updated to use this

src/utils/build_importer.py
maximo1491: Potentially, I've not really used XPath before, I did give it a try but couldn't get it to work so just did it this way. Not ideal, but at least it's clear what it's doing and doesn't need to be performant here.

aeon0:
That one didnt work properly:
['lucky_hit:_up_to_a_chance_to_make_enemies_vulnerable_for_seconds']

config/params.ini
```diff
@@ -14,6 +14,8 @@ hidden_transparency=0.35
 ; In case you get a warning about the LocalPref file not being found, you can either ignore the warning,
 ; or specify the correct path for your system here
 local_prefs_path=
+; Which browser to use to get builds, Chrome, Edge, or Firefox are currently supported
+browser=Edge
```
chrisoro:
```suggestion
browser=Chrome
```
Nitpick: Chrome seems to be the most used, so set the default to it

src/config/models.py
```diff
@@ -153,6 +153,7 @@ class ColorsModel(_IniBaseModel):
 
 
 class GeneralModel(_IniBaseModel):
+    browser: str = "Edge"
```
chrisoro: Please use a str enum here. See  keep_aspects as a reference

src/config/models.py
```diff
@@ -161,6 +162,12 @@ class GeneralModel(_IniBaseModel):
     profiles: list[str]
     run_vision_mode_on_startup: bool
 
+    @field_validator("browser")
```
chrisoro: When changing the type from str to enum, you can delete this validator

src/utils/build_importer.py
```diff
@@ -7,66 +10,95 @@
...
+ITEM_MODIFIERS_CLASS = "d4t-item-mods"
+ITEM_RARITY_CLASS = "d4t-item-options"
+ITEM_NAME_CLASS = "d4t-header-title"
 
 def import_build():
     Logger.info(f"Paste maxroll.gg build guide or planner build here ie https://maxroll.gg/d4/build-guides/minion-necromancer-guide")
     url = input()
-    if "maxroll.gg/d4/planner" not in url and "maxroll.gg/d4/build-guides/" not in url:
+    if MAXROLL_PLANNER_URL not in url and MAXROLL_BUILD_GUIDE_URL not in url:
         Logger.info(f"Invalid url, please use a d4 planner build on MaxRoll.gg")
```
chrisoro:
```suggestion
        Logger.error(f"Invalid url, please use a d4 planner build on MaxRoll.gg")
```

src/utils/build_importer.py
```diff
@@ -7,66 +10,95 @@
...
         return
-    options = webdriver.ChromeOptions()
-    options.add_argument("--headless=new")
-    driver = webdriver.Chrome(options=options)
+    match IniConfigLoader().general.browser:
```
chrisoro, line 40: When changing the type to enum, you can simplify this a lot. Also, no need for a catch all at the end since the config is validated first and would never reach this code with an invalid value

chrisoro (changes requested):
Even better than before. Just one more tiny thing :)

config/params.ini
maximo1491: Would agree, but I'd argue most people are running on Windows, so should always have Edge installed. I myself use Firefox over Chrome, but I'm never going to die on a hill defending Edge, so I'll make this change :D

Maximo1491:
Thanks so much for taking the time to review this. I am a c++ programmer by trade who only really dabbles with Python, its great to see how proper projects setup their code in Python!

config/params.ini
chrisoro: I guess you are right. This tool is only working on Windows anyways and all of them have Edge. Either way, all good :)

chrisoro (approved):

PR #233, opened 2024-05-20
Use ruff for formatting / linting

chrisoro:
:)


pyproject.toml
```diff
@@ -1,2 +1,51 @@
...
 line-length = 140
+target-version = "py312"
+
+[tool.ruff.format]
+docstring-code-format = true
+docstring-code-line-length = "dynamic"
+indent-style = "space"
+line-ending = "auto"
+quote-style = "double"
+skip-magic-trailing-comma = true
```
johannes-do: I would like to have the option to decide by the trailing comma if I want to split things in seperate lines or not. Setting this to false and split-on-trailing-comma = true does the trick.

pyproject.toml
```diff
@@ -1,2 +1,51 @@
...
+    "W",
+    "YTT",
+    # "PL",
+    # "PTH",
+    # "RUF",
+]
+ignore = ["E501"]
+
+[tool.ruff.lint.isort]
+split-on-trailing-comma = false
```
johannes-do: Set to true for reason stated in L12

src/item/descr/texture.py
```diff
@@ -44,14 +41,7 @@ def _find_bullets(
...
     )
-    small_bullets = search(
-        ref=template_list,
-        inp_img=img_item_descr,
-        threshold=threshold,
-        roi=roi_bullets,
-        use_grayscale=True,
-        mode=mode,
-    )
+    small_bullets = search(ref=template_list, inp_img=img_item_descr, threshold=threshold, roi=roi_bullets, use_grayscale=True, mode=mode)
```
johannes-do: revert to split by trailing comma

aeon0 (changes requested):
I am convinced. This is great!

PR #234, opened 2024-05-21
correct order of hotkeys in main

aeon0:


src/item/descr/texture.py
```diff
@@ -20,6 +20,8 @@ def _filter_outliers(template_matches: list[TemplateMatch]) -> list[TemplateMatc
     # Extract center[0] values
     centers_x = [tm.center[0] for tm in template_matches]
     # Calculate the median
+    if len(centers_x) == 0:
```
chrisoro:
```suggestion
    if not centers_x:
```

assets/lang/enUS/affixes.json
BimmerHex, line 1: Sorry, I don't know exactly how to use this place, but the line below in affixes needs to be deleted. There is a typo. Frankly, it already exists.

"damage_whiile_in_human_form": "damage whiile in human form",

PR #238, opened 2024-05-22
Fix for bug

Maximo1491:
5.1.1 import taking tempered stats instead of main stats


chrisoro:
As discussed, refactored the logic to be more robust here: https://github.com/aeon0/d4lf/pull/239

PR #261, opened 2024-06-01
Added ability to move items from or to the stash by pressing F7/F8 by…

cjshrader:
… default


src/config/loader.py
```diff
@@ -48,6 +48,8 @@ def _load_params(self):
             run_filter=self._select_val("advanced_options", "run_filter"),
             run_scripts=self._select_val("advanced_options", "run_scripts"),
             scripts=self._select_val("advanced_options", "scripts").split(","),
+            move_to_inv=self._select_val("advanced_options", "move_to_inv"),
```
chrisoro: As discussed, please keep it sorted alphabetically.

src/config/loader.py
```diff
@@ -59,6 +61,7 @@ def _load_params(self):
             keep_aspects=self._select_val("general", "keep_aspects"),
             profiles=self._select_val("general", "profiles").split(","),
             run_vision_mode_on_startup=self._select_val("general", "run_vision_mode_on_startup"),
+            move_item_type=self._select_val("general", "move_item_type"),
```
chrisoro: As discussed, please keep it sorted alphabetically.

src/config/models.py
```diff
@@ -24,6 +24,12 @@ class HandleRaresType(enum.StrEnum):
     junk = enum.auto()
 
 
+class MoveItemsType(enum.StrEnum):
+    junk = enum.auto()
```
chrisoro, line 29: As discussed, please keep it sorted alphabetically.

src/config/models.py
```diff
@@ -116,16 +122,18 @@ class AdvancedOptionsModel(_IniBaseModel):
     process_name: str = "Diablo IV.exe"
     run_filter: str
     run_scripts: str
+    move_to_inv: str
```
chrisoro: As discussed, please keep it sorted alphabetically.

src/config/models.py
```diff
@@ -116,16 +122,18 @@ class AdvancedOptionsModel(_IniBaseModel):
...
+    move_to_chest: str
     scripts: list[str]
 
     @model_validator(mode="after")
     def key_must_be_unique(self) -> "AdvancedOptionsModel":
-        keys = [self.exit_key, self.run_filter, self.run_scripts]
+        keys = [self.exit_key, self.run_filter, self.run_scripts, self.move_to_chest, self.move_to_inv]
```
chrisoro: As discussed, please keep it sorted alphabetically.

src/config/models.py
```diff
@@ -116,16 +122,18 @@ class AdvancedOptionsModel(_IniBaseModel):
...
         if len(set(keys)) != len(keys):
             raise ValueError("hotkeys must be unique")
         return self
 
-    @field_validator("exit_key", "run_scripts", "run_filter")
+    @field_validator("exit_key", "run_scripts", "run_filter", "move_to_chest", "move_to_inv")
```
chrisoro: As discussed, please keep it sorted alphabetically.

src/config/models.py
```diff
@@ -172,6 +180,7 @@ class GeneralModel(_IniBaseModel):
     language: str = "enUS"
     profiles: list[str]
     run_vision_mode_on_startup: bool
+    move_item_type: MoveItemsType = MoveItemsType.non_favorites
```
chrisoro: As discussed, please keep it sorted alphabetically.

src/overlay.py
```diff
@@ -17,6 +19,12 @@
 lock = threading.Lock()
 
 
+class FilterOrMoveOption(enum.StrEnum):
+    filter = enum.auto()
```
chrisoro: As discussed, please keep it sorted alphabetically.

chrisoro (changes requested):
Looks great! Just some minor stuff. I will do a full review tomorrow, but no big changes are needed.

src/overlay.py
```diff
@@ -133,37 +141,44 @@ def toggle_size(self):
         win_spec = WindowSpec(IniConfigLoader().advanced_options.process_name)
         move_window_to_foreground(win_spec)
 
-    def filter_items(self):
+    def filter_items(self, filter_or_move: FilterOrMoveOption = FilterOrMoveOption.filter):
```
chrisoro: there might be more additions in the future, so I would like to not add everything into this method. please extract the "move" functionality and add a corresponding method. generalize common stuff into a helper to reduce code duplications between your new move and this filter method

src/overlay.py
```diff
@@ -133,37 +141,44 @@ def toggle_size(self):
...
+                    self.loot_filter_thread = threading.Thread(target=self._wrapper_run_loot_filter, args=(filter_or_move,), daemon=True)
                     self.loot_filter_thread.start()
                     self.filter_button.config(fg="#006600")
             finally:
                 lock.release()
         else:
             return
 
-    def _wrapper_run_loot_filter(self):
+    def _wrapper_run_loot_filter(self, filter_or_move: FilterOrMoveOption = FilterOrMoveOption.filter):
```
chrisoro: this wrapper should be generalized so it takes a function as arg. RIght now, we have to differentiate on multiple levels. I would structure it differently:

- different methods for different hotkey
- they all do their stuff and call this thing in a new thread with a specific sub function. this "wrapper" can be general

src/loot_mover.py
```diff
@@ -0,0 +1,92 @@
...
+from src.utils.custom_mouse import mouse
+
+
+def move_items_to_stash():
+    Logger().info("Move inventory items to stash")
+
+    inv = CharInventory()
+    chest = Chest()
+
+    if chest.is_open():
```
chrisoro: I like to handle the bad case as soon as possible so we don't have so many open blocks. In your case, you can check if the chest is not open, then log and return.

src/loot_mover.py
```diff
@@ -0,0 +1,92 @@
...
+        occupied_inv, _ = inv.get_item_slots()
+
+        for i in IniConfigLoader().general.check_chest_tabs:
+            chest.switch_to_tab(i)
+
+            _, empty_chest = chest.get_item_slots()
+
+            if len(empty_chest) == 0:
```
chrisoro:
```suggestion
            if not empty_chest:
```

src/loot_mover.py
```diff
@@ -0,0 +1,92 @@
...
+        Logger().error("Moving items only works if stash is open")
+
+
+def move_items_to_inventory():
+    Logger().info("Move stash items to inventory")
+
+    inv = CharInventory()
+    chest = Chest()
+
+    if chest.is_open():
```
chrisoro: check bad case first to remove open blocks

src/loot_mover.py
```diff
@@ -0,0 +1,92 @@
...
+            if empty_slot_count < 1:
+                break
+
+        mouse.move(*Cam().abs_window_to_monitor((0, 0)))
+        Logger().info("Completed move")
+    else:
+        Logger().error("Moving items only works if stash is open")
+
+
+def move_items(inv, occupied, num_to_move):
```
chrisoro: I don't know if this was intentional or not, but some types are passed by reference in Python, and others are not. In your case, you are changing the list of occupied. This makes it very hard to understand when logic gets more complicated. I would recommend refactoring this to avoid doing that.

src/loot_mover.py
```diff
@@ -0,0 +1,92 @@
...
+            chest.switch_to_tab(i)
+
+            _, empty_chest = chest.get_item_slots()
+
+            if len(empty_chest) == 0:
+                continue
+
+            move_items(inv, occupied_inv, len(empty_chest))
+
+            if len(occupied_inv) == 0:
```
chrisoro: See my other comment on move_items.

chrisoro (changes requested):
Also, please add the new feature to readme

src/loot_mover.py
cjshrader: It was intentional to modify occupied, but I can have it return a tuple instead if that makes more sense.

src/overlay.py
cjshrader: This makes sense, let me know if I developed this in the way you envisioned

src/overlay.py
cjshrader: It is apparently very easy to pass methods in python, thank you for this tip

src/loot_mover.py
cjshrader: I made this change. Some of the younger devs on my team also handle things like this, I think the standards have just changed on me in the last 17 years and I need to get used to it.

src/loot_mover.py
cjshrader: I've made the change I mentioned above, let me know if that's what you were thinking.

chrisoro (dismissed):

src/loot_mover.py
chrisoro: Nice!

src/loot_mover.py
chrisoro: :)

src/overlay.py
chrisoro: Much better, Thank you for cleaning up our old stuff

PR #267, opened 2024-06-04
Added ability to do a "force recheck" which will reset all junk/favor…

cjshrader:
…ite items in whatever is being checked.

### DIFF
--- a/src/config/helper.py
+++ b/src/config/helper.py
@@ -13,8 +13,10 @@ def check_greater_than_zero(v: int) -> int:
 
 
 def key_must_exist(k: str) -> str:
-    if k not in keyboard._winkeyboard.from_name:
-        raise ValueError("key does not exist")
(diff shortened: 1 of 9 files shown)


src/config/helper.py
```diff
@@ -13,8 +13,10 @@ def check_greater_than_zero(v: int) -> int:
 
 
 def key_must_exist(k: str) -> str:
-    if k not in keyboard._winkeyboard.from_name:
-        raise ValueError("key does not exist")
+    all_keys = k.split("+")  # Handles modifiers like shift
```
chrisoro: Add check for valid modifier. Probably also rename method to show the new purpose

chrisoro (changes requested):
readme must be touched as well

chrisoro (approved):

PR #292, opened 2024-06-14
Refactored button overlay

suchyindustries:
- now ruff checks passed
- Refactored Logging

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -33,3 +33,4 @@ playground.py
 stats/
 utils/live-view/
 venv
+*backup*
--- a/requirements.txt
+++ b/requirements.txt
@@ -4,6 +4,7 @@ beautifultable
 colorama
 coverage
 cryptography
+httpx
 keyboard
 lxml
 mouse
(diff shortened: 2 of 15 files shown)


cjshrader:
Hey, I think a couple of changes might be needed here.

image

First off, after first click the Filter button turns into a dark grey that is hard to see. When vision is off it's pretty hard to see too.

Secondly, I wouldn't add a button for force refresh. I think people are going to have an expectation that when they click filter a filter is going to happen. I don't think that should be two clicks. I also think very very few people are going to use the refresh functionality, hiding it behind a hotkey seems fine to me.

If you make a hotkey something like ctrl+shift+f11 will this handle that length nicely?

This isn't specific to this PR, but I think if we're going to make the overlay bigger (which I'm not opposed to, I particularly like the larger console size) we need to do two things in the future:
- Add an option to completely disable the overlay for those who don't use it
- Figure out a way to make the overlay active only when D4 is the active window. At this size it obstructs my windows bar

PR #300, opened 2024-06-16
config_tab.py

suchyindustries:
now cancel button don't reset values


cjshrader:
Good catch, I thought those things returned true or false not what button was pressed. I learned something, thank you.

chrisoro (dismissed):
ty sir

PR #317, opened 2024-06-20
Added ability to refresh the status of all items without performing a…

cjshrader:
… filter after

### DIFF
--- a/src/config/models.py
+++ b/src/config/models.py
@@ -52,6 +52,11 @@ class ComparisonType(enum.StrEnum):
     smaller = enum.auto()
 
 
+class ForceRefreshType(enum.StrEnum):
+    with_filter = enum.auto()
+    without_filter = enum.auto()
+
+
 class _IniBaseModel(BaseModel):
     model_config = ConfigDict(extra="forbid", str_strip_whitespace=True, validate_assignment=True)
 
@@ -137,6 +142,11 @@ class AspectUniqueFilterModel(AffixAspectFilterModel):
 
 class AdvancedOptionsModel(_IniBaseModel):
     exit_key: str = Field(default="f12", description="Hotkey to exit d4lf", json_schema_extra={IS_HOTKEY_KEY: "True"})
+    force_refresh_only: str = Field(
+        default="ctrl+shift+f11",
+        description="Hotkey to refresh the junk/favorite status of all items in your inventory/stash. A filter is not run after.",
+        json_schema_extra={IS_HOTKEY_KEY: "True"},
+    )
     log_lvl: LogLevels = Field(default=LogLevels.info, description="The level at which logs are written")
     move_to_chest: str = Field(
         default="f8", description="Hotkey to move configured items from inventory to stash", json_schema_extra={IS_HOTKEY_KEY: "True"}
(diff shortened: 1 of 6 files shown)


src/overlay.py
```diff
@@ -33,6 +33,7 @@ def emit(self, record):
 
 class Overlay:
     def __init__(self):
+        LOGGER.setLevel(str(IniConfigLoader().advanced_options.log_lvl).upper())
```
chrisoro: Why do this? log level is already set in main and all other logger inherit the level

chrisoro (changes requested):

src/config/models.py
```diff
@@ -52,6 +52,11 @@ class ComparisonType(enum.StrEnum):
     smaller = enum.auto()
 
 
+class ForceRefreshType(enum.StrEnum):
```
chrisoro: Do you think we add new items to this enum? If not, remove it and change the arg inside the function to bool

src/gui/config_tab.py
```diff
@@ -420,7 +420,7 @@ def keyPressEvent(self, event):
         if non_mod_key in modifiers_str:
             non_mod_key = ""
 
-        key_str = " + ".join(modifiers_str + [non_mod_key])
+        key_str = "+".join(modifiers_str + [non_mod_key])
```
chrisoro, line 423: Was this bugget before with " + " or just cosmetic change?

src/overlay.py
cjshrader: For the overlay terminal specifically, the log level is being set to debug at all times. If you check it before this change you'll see it's 0.

Perhaps this is a band-aid fix for something deeper? I don't know, but I know this resolves it.

src/config/models.py
cjshrader: There are 3 options here: Do no force refresh (I use None for that), force refresh with no filter, force refresh with a filter.

Would it make more sense then for me to add that third option to the enum instead of using None?

src/gui/config_tab.py
cjshrader: There was no bug, but this is just consistent with how we set our defaults and takes up a tiny amount less space when we print the table in the console that lists the hotkeys.

src/config/models.py
cjshrader: Just to add - I doubt new options will ever be added.

chrisoro: Yes instead of magical value None, lets just add the third option here and replace the None checks

src/gui/config_tab.py
chrisoro: OK

src/overlay.py
chrisoro: Ok I see. Please change line 104 which sets the level of this handler to

```LOGGER.root.handlers[0].level```

of course guard it if no handlers are available and then just set it to 0

Comment (cjshrader) [reply src/overlay.py]: I made this change and it appears to work great now, I tested at debug/info/warning levels and all were working as expected.

File: src/config/models.py (hunk 2 above)
Comment (cjshrader) [reply src/config/models.py]: All done

File: src/overlay.py [hunk 4]
```diff
@@ -101,7 +103,8 @@ def __init__(self):
 
         # Setup the listbox logger handler
         textlog_handler = TextLogHandler(self.terminal_text)
-        textlog_handler.setLevel(LOGGER.level)
+        log_level = LOGGER.root.handlers[0].level if LOGGER.root.handlers[0] else 0
```
chrisoro:
```suggestion
        log_level = LOGGER.root.handlers[0].level if LOGGER.root.handlers else 0
```

chrisoro (approved):

PR #325, opened 2024-06-23
fix Eyes in The Dark unique description

rmobis:
not sure if/when this was updated, but it isn't working atm


chrisoro (approved):
Thank you sir. We generate the assets based on other data and have custom files to fix mistakes in the source data (d4data). I added your correction accordingly

PR #334, opened 2024-07-01
Add option to enable/disable mark as favorite

geocine:
Whether to favorite matched items or not. Defaults to true which uses the current behavior. If set to false, this will go pass matched items and will not mark them.


README.md
```diff
@@ -75,6 +75,7 @@ The config folder in `C:/Users/<WINDOWS_USER>/.d4lf` contains:
...
 | check_chest_tabs                                  | Which chest tabs will be checked and filtered for items in case chest is open when starting the filter. You need to buy all slots. Counting is done left to right. E.g. 1,2,4 will check tab 1, tab 2, tab 4                                                                         |
 | move_to_inv_item_type<br/>move_to_stash_item_type | Which types of items to move when using fast move functionality. Will only affect tabs defined in check_chest_tabs. <br>- `favorites`: Move favorites only <br>- `junk`: Move junk only <br>- `non_favorites`: Only items not marked as favorite <br>- `everything`: Move everything |
+| mark_as_favorite                                  | Whether to favorite matched items or not. Defaults to true. If set to false, this will go pass matched items and will not mark them                                                                                                                                                  |
```
chrisoro:
```suggestion
| mark_as_favorite                                  | Whether to favorite matched items or not. Defaults to true |
```

src/loot_filter.py
```diff
@@ -37,7 +37,7 @@ def check_items(inv: InventoryBase, force_refresh: ItemRefreshType):
     start_time = None
     img = None
     for item in occupied:
-        if item.is_junk or item.is_fav:
+        if item.is_junk or (item.is_fav and IniConfigLoader().general.mark_as_favorite):
```
chrisoro: Can't we still skip items that are either junk or faved already?

src/loot_mover.py
```diff
@@ -98,7 +98,9 @@ def _move_items(
         if (
             (move_item_type == MoveItemsType.favorites and item.is_fav)
             or (move_item_type == MoveItemsType.junk and item.is_junk)
-            or (move_item_type == MoveItemsType.non_favorites and not item.is_fav)
+            or (move_item_type == MoveItemsType.non_favorites and (
```
chrisoro: Why do we need to change the loot mover?

chrisoro (changes requested):

README.md
geocine: ~~What has changed here? I guess white spaces?~~ I was on mobile and did not see the entire change.

src/loot_filter.py
geocine: True, I can revert that change, since we already have a reset.

src/loot_mover.py
geocine: Right now `non_favorite` items also move `junk` which makes sense if there are only 2 states: `junk` and `favorite` . When `mark_as_favorite` is set to `false` there will be 3 states: `junk`, `favorite` and `unmarked` . `unmarked` is both not favorite and not junk. At this point, this is how I expect `non_favorite` to work since we already have separate options for `favorite` and `junk`. Another suggestion to avoid this complexity and confusion is to just rename `non_favorite` to `not_marked` . Because right now `non_favorite` also means `junk`. I hope this makes sense.

cjshrader: I think I get it, but would it be more clear to just add an "unmarked" option to the loot mover instead? I don't think I love the idea of it working different ways depending on what's set for a completely unrelated property.

chrisoro: +1 for just adding all possibilities as is:
- junk
- favorites
- plain
- all

geocine: thanks both, I will take a stab at this. On the code I can see we have 4 existing options.

-  everything
-  favorites
-  junk
-  non_favorites

I am planning to change `non_favorites` to `unmarked`  and change the behavior all together. what do you think? Also set defaults for loot mover to `everything` as its the neutral option in my opinion

README.md
```diff
@@ -79,6 +79,7 @@ The config folder in `C:/Users/<WINDOWS_USER>/.d4lf` contains:
 | run_vision_mode_on_startup                        | If the vision mode should automatically start when starting d4lf. Otherwise has to be started manually with the vision button or the hotkey                                                                                                                                          |
 | check_chest_tabs                                  | Which chest tabs will be checked and filtered for items in case chest is open when starting the filter. You need to buy all slots. Counting is done left to right. E.g. 1,2,4 will check tab 1, tab 2, tab 4                                                                         |
 | move_to_inv_item_type<br/>move_to_stash_item_type | Which types of items to move when using fast move functionality. Will only affect tabs defined in check_chest_tabs. <br>- `favorites`: Move favorites only <br>- `junk`: Move junk only <br>- `non_favorites`: Only items not marked as favorite <br>- `everything`: Move everything |
```
chrisoro: needs to be updated

chrisoro (changes requested):
looks good. please update the readme

chrisoro (approved):

PR #335, opened 2024-07-01
Update build.py

suchyindustries:
As d4lf doesn't update changes "on the fly", I think everybody's behaviour is: after setting up in GUI, start the main program. 
So this is my proposed change:
From now on, after "d4lf.exe --gui" closes, it will automatically start the main d4lf.exe. Try it yourself with this batch script:

@echo off
cd /d "%~dp0"
d4lf.exe --gui
start "" d4lf.exe


cjshrader:
I don't think this really matches how I personally use d4lf. I keep the GUI up a lot even if I don't have d4lf running. I wouldn't want d4lf to automatically run when I close it, especially since we have nothing in place at this time to prevent multiple copies of d4lf from running.

I have a suggestion post in discord for ways to handle stopping/starting d4lf and I'd love your feedback there. I haven't been able to work on it much at all lately.

PR #348, opened 2024-08-11
Refactored tempered detection

chrisoro:


cjshrader (approved):
Nice work, you're so fast at this stuff

PR #351, opened 2024-08-11
Update custom_uniques_enUS.json

tikalein:
Crown of Lucion addedd


chrisoro (changes requested):
We generate this file. Please also edit the source:
src/tools/data/custom_uniques_enUS.json

chrisoro (approved):
oops my bad. all good, thank you!

PR #352, opened 2024-08-12
Added default minpower to importers

cjshrader:


chrisoro:
I get why you added this but the value makes 0 sense and is now always annoying. I would say if you want a template to edit for later, just enable full_dump. What do you think?

chrisoro:
Sure we can ask but I think minPower is super irrelevant. For leveling, it doesn't matter since it is easy anyways. For endgame you only really care for weapon ilvl. For everything else the stats are most important. And then you are at a point where only 925 frops anyways 🤷

cjshrader:
I just don't see what it hurts to have it, and it gives a minor improvement to people especially while leveling.

chrisoro (approved):

PR #374, opened 2024-08-27
Added ability to configure how uniques that do not match a filter are handled, and additionally added unique importing to importers

cjshrader:
Uniques are now handled in two ways that I think make sense in practice but is hard to explain.

Let's say we have a tibault's will with 900 item power.

If we only have this in our profiles:

```
Uniques:
- aspect: {name: tibaults_will}
  minPower: 925
```

Then naturally the Tibault's marked as junk. If we change minPower to 875, it will be favorited. That functionality has not changed.

If we also have a Crown of Lucion in our inventory, it matches no unique filters. What happens to it now is configurable through the new handle_uniques property, which defaults to favorite.

If instead our filter looks like this:

```
Uniques:
- aspect: {name: tibaults_will}
  minPower: 925
- itemType: [pants]
```
Now there is a filter that applies to everything as itemType is a global filter. Our Tibault's Will will be marked as favorite because it matches the second filter (the item type one). Crown of Lucion will be marked as junk because it matches neither filter. handle_uniques no longer applies because there is now a filter that applies to every single item.

What this has allowed me to do is add unique importing to the importers because having aspect-only filters now doesn't junk uniques outside of that aspect with the default configurations. I don't think people will use the global filters very often, but they continue to be supported if needed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,19 +72,20 @@ The config folder in `C:/Users/<WINDOWS_USER>/.d4lf` contains:
 
 ### params.ini
 
-| [general]                                         | Description                                                                                                                                                                                                                                                                                                                  |
-|---------------------------------------------------|------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
-| profiles                                          | A set of profiles separated by comma. d4lf will look for these yaml files in config/profiles and in C:/Users/WINDOWS_USER/.d4lf/profiles                                                                                                                                                                                     |
-| keep_aspects                                      | - `all`: Keep all legendary items <br>- `upgrade`: Keep all legendary items that upgrade your codex of power <br>- `none`: Keep no legendary items based on aspect (they are still filtered!)                                                                                                                                |
-| handle_rares                                      | - `filter`: Filter them based on your profiles <br>- `ignore`: Ignores all rares, vision mode shows them as blue and auto mode never junks or favorites them <br>- `junk`: Vision mode shows them always as red, auto mode always junks rares                                                                                |
-| run_vision_mode_on_startup                        | If the vision mode should automatically start when starting d4lf. Otherwise has to be started manually with the vision button or the hotkey                                                                                                                                                                                  |
-| check_chest_tabs                                  | Which chest tabs will be checked and filtered for items in case chest is open when starting the filter. You need to buy all slots. Counting is done left to right. E.g. 1,2,4 will check tab 1, tab 2, tab 4                                                                                                                 |
-| move_to_inv_item_type<br/>move_to_stash_item_type | Which types of items to move when using fast move functionality. Will only affect tabs defined in check_chest_tabs. You can select more than one option. <br>- `favorites`: Move favorites only <br>- `junk`: Move junk only <br>- `unmarked`: Only items not marked as favorite or junk <br>- `everything`: Move everything |
-| mark_as_favorite                                  | Whether to favorite matched items or not. Defaults to true                                                                                                                                                                                                                                                                   |
-| minimum_overlay_font_size                         | The minimum font size for the vision overlay, specifically the green text that shows which filter(s) are matching. Note: For small profile names, the font may actually be larger than this size but will never go below this size.                                                                                          |
-| hidden_transparency                               | The overlay will become transparent after not hovering it for a while. This can be changed by specifying any value between [0, 1] with 0 being completely invisible and 1 completely visible                                                                                                                                 |
-| browser                                           | Which browser to use to get builds, please make sure you pick an installed browser: chrome, edge or firefox are currently supported                                                                                                                                                                                          |
-| full_dump                                         | When using the import build feature, whether to use the full dump (e.g. contains all filter items) or not                                                                                                                                                                                                                    |
+| [general]                                         | Description                                                                                                                                                                                                                                                                                                                                                                                                                              |
+|---------------------------------------------------|------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| profiles                                          | A set of profiles separated by comma. d4lf will look for these yaml files in config/profiles and in C:/Users/WINDOWS_USER/.d4lf/profiles                                                                                                                                                                                                                                                                                                 |
+| keep_aspects                                      | - `all`: Keep all legendary items <br>- `upgrade`: Keep all legendary items that upgrade your codex of power <br>- `none`: Keep no legendary items based on aspect (they are still filtered!)                                                                                                                                                                                                                                            |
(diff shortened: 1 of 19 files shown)


README.md
```diff
@@ -72,19 +72,20 @@ The config folder in `C:/Users/<WINDOWS_USER>/.d4lf` contains:
...
+| handle_rares                                      | - `filter`: Filter them based on your profiles <br>- `ignore`: Ignores all rares, vision mode shows them as blue and auto mode never junks or favorites them <br>- `junk`: Vision mode shows them always as red, auto mode always junks rares                                                                                                                                                                                            |
+| handle_uniques                                    | How to handle uniques that do not match any filter. This property does not apply to filtered uniques. All mythics are favorited regardless of filter. <br/>- `favorite`: Mark the unique as favorite and vision mode will show it as green (default)<br/>- `ignore`: Do nothing with the unique and vision mode will show it as green<br/>- `junk`: Mark any uniques that don't match any filters as junk and show as red in vision mode |
```
chrisoro, line 80:
```suggestion
| handle_uniques                                    | How to handle uniques that do not match any filter. This property does not apply to filtered uniques. All mythics are favorited regardless of filter. <br/>- `favorite`: Mark the unique as favorite and vision mode will show it as green (default)<br/>- `ignore`: Do nothing with the unique and vision mode will show it as green<br/>- `junk`: Mark the unique as junk and show as red in vision mode |
```
Already mentioned in the first sentence that you are only talking about uniques that don't match

README.md
```diff
@@ -340,9 +341,21 @@ names in [assets/lang/enUS/sigils.json](assets/lang/enUS/sigils.json).
 ### Uniques
 
 Uniques are defined by the top-level key `Uniques`. It contains a list of parameters that you want to filter for. If no
-Unique filter is provided, all unique items will be kept.
-Uniques can be filtered similar to [Affixes](#Affixes) but due to their nature of fixed
-effects, you only have to specify the thresholds that you want to apply.
+Unique filter is provided, uniques will be handled according to the handle_uniques configuration. All mythics are
```
chrisoro, line 344:
```suggestion
Unique filter is provided, uniques will be handled according to the handle_uniques configuration. All mythic uniques are 
```

requirements.txt
```diff
@@ -28,6 +28,7 @@ pytweening
 pywin32
 pyyaml
 rapidfuzz
+ruamel.yaml==0.18.6
```
chrisoro, line 31: Why can't we use pyyaml?

src/config/models.py
```diff
@@ -444,9 +457,10 @@ def min_max_tier_in_range(cls, v: int) -> int:
 
 class UniqueModel(BaseModel):
     model_config = ConfigDict(extra="forbid")
-    affix: list[AffixFilterModel] = []
     aspect: AspectUniqueFilterModel = None
+    affix: list[AffixFilterModel] = []
```
chrisoro, line 461: keep it alphabetical please

src/gui/importer/d4builds.py
```diff
@@ -104,6 +107,19 @@ def import_d4builds(driver: ChromiumDriver = None, url: str = None):
...
+
+        if slot_to_unique_name_map[slot]:
+            unique_model = UniqueModel()
+            unique_name = slot_to_unique_name_map[slot]
+            try:
+                unique_model.aspect = AspectUniqueFilterModel(name=unique_name)
+                unique_model.affix = [AffixFilterModel(name=x.name) for x in affixes]
+                unique_filters.append(unique_model)
+            except Exception as e:
+                print(traceback.format_exc())
```
chrisoro: Please use LOGGER.exception instead of .error. It will show the full callstack/traceback. Can be removed here then.

src/gui/importer/maxroll.py
```diff
@@ -51,12 +52,24 @@ def import_maxroll(url: str):
...
+            unique_name = mapping_data["items"][resolved_item["id"]]["name"]
+            try:
+                unique_model.aspect = AspectUniqueFilterModel(name=unique_name)
+                unique_model.affix = [
+                    AffixFilterModel(name=x.name)
+                    for x in _find_item_affixes(mapping_data=mapping_data, item_affixes=resolved_item["explicits"])
+                ]
+                unique_filters.append(unique_model)
+            except Exception as e:
+                print(traceback.format_exc())
```
chrisoro: same as in d4builds

src/gui/importer/mobalytics.py
```diff
@@ -109,6 +113,19 @@ def import_mobalytics(url: str):
...
+        if unique_name:
+            unique_model = UniqueModel()
+            try:
+                unique_model.aspect = AspectUniqueFilterModel(name=unique_name)
+                if affixes:
+                    unique_model.affix = [AffixFilterModel(name=x.name) for x in affixes]
+                unique_filters.append(unique_model)
+            except Exception as e:
+                print(traceback.format_exc())
+                LOGGER.error(f"Unexpected error importing unique {unique_name}, please report a bug: {str(e)}")
```
chrisoro: same as in d4builds

chrisoro (changes requested):
Good stuff, only minor remarks.
Can you please create some unittests on a filter level? See `tests/item/filter/filter_test.py`

requirements.txt
cjshrader: pyyaml converts the model to a JSON string and then uses json.loads to read the string. This will sort the keys alphabetically which is not what we want. Users are used to everything looking like this:

```
Uniques:
  - aspect: [ tibaults_will, 25 ]
    affix:
      - { name: damage_reduction_from_close_enemies, value: 12 }
```

With the default from pyyaml, it is instead printed like this:

```
Uniques:
-  affix:
      - { name: damage_reduction_from_close_enemies, value: 12 }
   aspect: [ tibaults_will, 25 ]
```

This still works just fine for filtering, but I think people will find it confusing and we'll get endless questions about it. It especially looks confusing when you have multiple uniques there, which is what happens after importing.

src/config/models.py
cjshrader: See above comment, we want to keep aspect on top here in this specific case

src/gui/importer/d4builds.py
cjshrader: Can you confirm I used LOGGER.exception properly in my code? It looks like you don't pass in the exception or anything, is that correct? I tried to stay consistent with other usages but that seemed weird to me.

cjshrader:
@chrisoro when it comes to unit tests, I actually added a couple but I realized that my changes don't change any of the existing tests. In every instance where we have one item and one filter, the output will always be the same as it was previously.

The changes are when we have multiple items or multiple filters. Do you want me to add some tests around that? I don't know python unit testing too well but I can attempt something.

chrisoro:
Yes, exactly. Have two filters with uniques and check the new logic. Don't waste too much time. I can add some too over the weekend.

cjshrader:
Ok @chrisoro I wrote some unit tests. I also reworked should_keep and item_filter/vision_mode so that should_keep always returns what you expect (ie, if an item will be kept it always returns true. It didn't actually do this before, the logic was sometimes handled in item_filter/vision_mode)

chrisoro (dismissed):

chrisoro (approved):

src/gui/importer/d4builds.py
chrisoro: yes

PR #378, opened 2024-09-08
Read rarity from item description instead of border image if we can find it

cjshrader:


src/item/descr/item_type.py
```diff
@@ -108,3 +113,10 @@ def _find_sigil_tier(concatenated_str: str) -> int:
             if following_word.isdigit():
                 return int(following_word)
     return None
+
+
+def _find_item_rarity(concatenated_str: str) -> ItemRarity | None:
```
chrisoro, line 118: do fuzzy search so OCR artifacts are not as bad.
e.g. legendari should still match legendary (don't know if we already use it here, maybe when we match the others strings. else levenshtein is a good algorithm for that)

cjshrader: My concern is we'd end up matching some hypothetical item that might not exist. Like "Mage's staff" might fuzzy match "magic" and we wouldn't want that.

Right now, if it can't find the word, it'll just continue to use whatever the image search returned in the top left corner, which has been pretty reliable for a lot of people anyway. I was thinking if we end up with a "legendari" we could add it to corrections.json too.

I really don't know much about fuzzy search, if you still think it's a good idea I can look into it. One thing I am seeing if I can do is rework read_descr_season_5_test to see if all of the rarities are being read correctly and it's not just defaulting to the one we provide.

chrisoro: You are right, lets see how it goes first before making it more complicated.

chrisoro (approved):

PR #412, opened 2025-01-25
Added use_tts to deprecated keys so that people who are swapping back…

cjshrader:
… and forth don't need to manually remove it. This is only meant to be merged in to the main non-TTS branch.


chrisoro (approved):
good idea

PR #414, opened 2025-01-26
Add profile editor gui

Shebour:
I added a simple profile editor in the GUI as a new tab:

profile_editor_gui

- You can only edit an already downloaded profile configuration.
- You cannot add or delete items from the profile. It will be added in a future update if this one is accepted.
- The default profile loaded is the first one in the params.ini file.
- You can open a specific profile file to edit it.
- You can set minPower, minCount, minGreaterAffixCount, affixes and inherent affixes.


cjshrader:
I am testing this out. Some notes:

Please make the drop down boxes ignore the scroll wheel. You can reuse what I did in the config tab:

https://github.com/d4lfteam/d4lf/blob/d890d4a854651d2c7e18a5243b0702b0cc88efa9/src/gui/config_tab.py#L206

Please add a reset button to reset back to the values in the file, if that is not too difficult.

Otherwise I think this is a great starter point, I already used it on some of my profiles. The set all * buttons are great.

src/gui/build_from_yaml.py
```diff
@@ -0,0 +1,151 @@
+import yaml
```
cjshrader: Why isn't this using the existing ProfileModel in models.py? This seems like reinventing the wheel. It is not writing uniques out properly and I realize because it is not using the same method that the importers use to write the profile.

You can see a ProfileModel in use here:

https://github.com/d4lfteam/d4lf/blob/d890d4a854651d2c7e18a5243b0702b0cc88efa9/src/gui/importer/maxroll.py#L104

And here is where we save the file: https://github.com/d4lfteam/d4lf/blob/d890d4a854651d2c7e18a5243b0702b0cc88efa9/src/gui/importer/common.py#L170

src/gui/profile_tab.py
```diff
@@ -0,0 +1,210 @@
...
+        self.main_layout.addWidget(instructions_label)
+
+        instructions_text = QTextBrowser()
+        instructions_text.append("The default profile loaded is the first profile in the params.ini file. You can change the default profile in the params.ini file by changing the 'profiles' value to the desired profile name.")
+        instructions_text.append("You can also load a profile by clicking the 'File' button.")
+        instructions_text.append("")
+        instructions_text.append("All values are not saved automatically immediately upon changing.")
+        instructions_text.append("You must click the save button to apply the changes to the profile.")
+        instructions_text.append("")
+        instructions_text.append("Note: You will need to restart d4lf after modifying these values. Modifying profile file manually while this gui is running is not supported (and really not necessary).")
```
cjshrader: You do not need to restart d4lf for profile changes. It picks those up automatically

src/gui/build_from_yaml.py
Shebour: How can I know which importer to use ? (Maxroll / diablo trade / d4builds)

cjshrader: In the end they all write it out more or less the same way, and I think you can reuse the functionality they use. My second link is in common.py

src/gui/profile_tab.py
Shebour: It does if I use the ProfileModel ? Because if I use my model it doesn't.

src/gui/build_from_yaml.py
Shebour: Can I load a yaml file with ProfileModel or do I need some parsing before to create the ProfileModel ? Because I'm loading the yaml profile not the url here.

cjshrader: I'm not sure if you deleted a comment or github is acting up, but check out filters.py load_files to see how we load the ProfileModel. It's directly from the yamls.

Shebour: Yes I deleted, at first I didn't found how to load the yaml directly but now it's all good, I can load the yaml in the ProfileModel directly. I'm gonna continue to adapt the code during this week.

Shebour:
- Using ProfileModel to load/save profile yaml
- Refresh button to reset all values from file
- Create / Delete items
- Fixing save alert when clicking Cancel, do not quit app.
- Using IgnoreScrollWheelComboBox and IgnoreScrollWheelSpinBox for SpinBox
https://github.com/Shebour/d4lf_main/blob/e3fe8f50e1168f5951c415af2e9a034620adde9b/src/gui/dialog.py#L8

cjshrader:
This is looking good, I'm going to make some small tweaks to it in the main branch and make a release

cjshrader (approved):

cjshrader:
There are a lot of formatting changes needed. We use a tool called ruff for that. For future development, please install ruff: https://github.com/astral-sh/ruff

I have it set up in Intellij IDEA to run every time I save a file, which happens automatically all the time. I don't know what IDE you use but I bet you can do the same.

For now, I'm fixing these issues myself in this pr: https://github.com/d4lfteam/d4lf/pull/423

cjshrader:
This was now merged and I'll make a release soon. I'll be closing this PR. Good work!

PR #460, opened 2025-04-26
Update draw_match_outline to dynamically change colors based on codex upgrade (#246)

### LINKED ISSUE #246: Request: Change vision filter codex upgrade color
Currently this shows green, the same as items matching whitelist. Would be nice if this was changed to yellow/orange or something to quickly ID it.

JoShMiQueL:
This pull request updates the `draw_match_outline` method in `vision_mode_with_highlighting.py` to dynamically change the highlight color based on whether the item has a codex upgrade.

Orange (`#fca503`) is used for items with a codex upgrade.
 Green (`#23fc5d`) is used otherwise.

This change improves visual differentiation for items.

Closes #246


cjshrader:
Thanks a bunch for this PR. A few notes:

First, I think you want to reverse the defaults. If it matches a profile AND it's an aspect upgrade, I think people care more that it matches a profile first. So it should default to green in that case. In my experience people only get confused when something is kept and it doesn't match any profiles, so a new color will be great there.

Secondly, what if they have keep_aspects set to None? If the item matches no profiles, the box should be red in that case. Is that the behavior with your change? I'd imagine so but it's worth double checking to be sure.

Third, should a third color be introduced for cosmetic upgrades? That could be a future update instead as well.

Finally, would you mind updating the keep_aspects section of the readme to mention the item is highlighted orange if it matches no profiles but is an aspect upgrade?

JoShMiQueL:
Thanks for your feedback 😁.

This commit should incorporate the change you proposed. https://github.com/d4lfteam/d4lf/pull/460/commits/6da4fbf6163b5c96a55324bb18cee7cbd5fefc1b

src/scripts/vision_mode_with_highlighting.py
```diff
@@ -207,6 +209,17 @@ def draw_no_match_outline(self, item_roi):
         self.root.update_idletasks()
         self.root.update()
 
+    def draw_codex_upgrade_outline(self, item_roi, item_descr):
+        x, y, w, h, off = self.get_coords_from_roi(item_roi)
+
+        self.create_signal_rect(self.canvas, w, self.thick, "#fca503")
+
+        # show string indicating that this item upgrades the codex
+        text_y = self.draw_text(self.canvas, "Codex Upgrade", "#fca503", h, 5, w // 2)
```
cjshrader: This line is failing test cases but is it even needed? I believe the profile checking section returns "Aspect Upgrade" so this text is already added. I'm not sure what happens in this case with this additional draw text

cjshrader: Not failing test cases, failing the pre-commit: https://github.com/d4lfteam/d4lf?tab=readme-ov-file#formatting--linting

cjshrader (changes requested):

src/scripts/vision_mode_with_highlighting.py
cjshrader: Looking at my earlier comment I understand now why this line was needed. Thank you for fixing the test case!

cjshrader (approved):

cjshrader:
Hey @JoShMiQueL do you see this working on your side? I have this item that's an Aspect upgrade but it's still green
image

PR #467, opened 2025-05-03
Profile Editor Update

Shebour:
- Affixes editor reworked for better readability
- Sigils, Tributes and Uniques editors added
- Check profile model validity before saving
- Fixed warning before closing without saving the profile
- Editor loaded when clicking on the tab to not overload the GUI when people don't want to edit their profile

Capture d'écran 2025-05-03 131310

Capture d'écran 2025-05-03 131340

Capture d'écran 2025-05-03 131423

Capture d'écran 2025-05-03 131442


cjshrader (approved):
Amazing work. I can't test out the actual functionality at this time, but I'm going to move forward with approval and we'll merge it in since most of the functionality is new it won't break anything if it doesn't work. When I finally get to make a character I'll definitely be using this.

Uniques are tricky and I'm really impressed you added them. That what I want to mess around with the most.

PR #473, opened 2025-05-04
Optimize tabs loading and fix profile loading cases

Shebour:
As you may have noticed, when clicking on "edit profile" you have to wait approximately 5 seconds for the tabs to load.
I delayed the loading of tab content when you click on the tab. Same for collapsible containers.
The content is loaded only the first time you click on the tab / collapsible, no unloading happen when you leave the tab so you won't have another loading when returning on the tab.

I also changed the profile loading, if you don't have any profile in the `params.ini`, the open file dialog will open only when you click on the tab "edit profile", if you cancel, it won't crash anymore and no profile will be loaded


cjshrader (approved):
This is looking great, I can't wait to get to use it

PR #548, opened 2025-07-03
add vision_mode_coordinates

chrisoro:
Dunno if you wanna add some doc or sensible defaults.. feel free to change accordingly


cjshrader:
Hmm I'm wondering what is easier for the user, coordinates or something like a percent of the screen.

I wish there was a way to not move its location by default but I can't think of anything, maybe if it's set to 0,0 we use the old method?

chrisoro:
can just make it optional in pydantic (allow None) and then check if its none and apply the old formula

PR #588, opened 2025-12-14
Some maxroll fixes

chrisoro:


assets/lang/enUS/uniques.json
```diff
@@ -48,7 +48,7 @@
         "num_inherents": 0
     },
     "banished_lords_talisman": {
-        "num_inherents": 0
+        "num_inherents": 1
```
cjshrader: Where are you pulling this data from? I have the latest from d4data and Diablo4Companion.

Amulets/rings don't have inherents anymore, because now they just have a flat "All resist" which we don't treat as an inherent just like we don't treat Armor as an inherent. So Banished Lord's Talisman and Dolmen Stone are correct to have 0.

Blue Rose looks different, it might actually have 1 inherent, I'm not sure, I'd need to see one in game. I didn't check the rest.

chrisoro: d4data has newest stuff on a branch. main is outdated (for now): https://github.com/DiabloTools/d4data/tree/2.5
d4companion just main. do we even look at these values?

cjshrader: Yes, these values are how we tell how many inherent a unique has when we're parsing it. I am using branch 2.5, that's why I'm confused where your data came from.

This example might be a little outdated, but if we look at Esu's Heirloom here is the TTS we'd get:

```
        [
            "ESUS HEIRLOOM",
            "Unique Boots",
            "750 Item Power",
            "60 Armor",
            "Evade Grants +100% Movement Speed for 3.0 Seconds [100]%[3.0]",
            "+17.0% Movement Speed [16.5 - 20.0]%",
            "+10.5% Movement Speed for 7 Seconds After Killing an Elite [9.0 - 11.0]%[7]",
            "+58.0% Critical Strike Damage [52.0 - 70.0]%",
            "+10.5% Resistance to All Elements [9.0 - 10.8]%",
            "Your Critical Strike Chance is increased by 34% [20 - 40]% of your Movement Speed bonus.. Current Bonus: 17.6% [10.4 - 20.7]%",
            "While scholars have proven these boots were not created by Esu herself, it is noteworthy that they have been passed down since the formation of the Mage Clans.. - Barretts Book of Implements",
            "Requires Level 60Sorcerer. Only. Unique Equipped",
            "Sell Value: 90,289 Gold",
            "Durability: 100/100",
            "Right mouse button",
        ]
```

Evade granting movement speed is an inherent, then there are 4 explicit affixes, then the aspect. We assume it has 4 explicit affixes but we have to know how many inherent affixes there are to know we've hit the aspect.

cjshrader: I just ran the gen_data again just to be sure (and made sure I had latest d4data and d4companion) and I'm not seeing the inherents get updated like this, I'm very confused.

chrisoro: That is strange. I generated on my Windows machine and it didn't result in any changes. Will try it again on mac, this is where I created this PR. Maybe there is some os dependency in our code,.

cjshrader: I know you added the case insensitivity to the files, which I appreciate, but I also don't see how they could be causing this.

chrisoro:
This has conflicting changes with your other pr. Do you still want to merge this? e.g. moving config options around.

I guess we merge this and then apply your changes

PR #589, opened 2025-12-15
Added max_stash_tabs property so that both 6 and 7 stash tab setups can be supported. Additionally, added fast_vision_mode_coordinates to advanced configs

cjshrader:


src/config/data.py
```diff
@@ -58,7 +58,8 @@
         slots_5x10=np.array([92, 538, 1224, 972]),
         sort_icon=np.array([2440, 1332, 126, 124]),
         stash_menu_icon=np.array([592, 144, 218, 96]),
-        tab_slots=np.array([240, 292, 930, 106]),
+        tab_slots_6=np.array([300, 284, 800, 102]),
```
chrisoro: imo bad design to hold both values in here, when the user configures just one. why not use one variable and lookup table?

src/config/models.py
```diff
@@ -311,6 +334,12 @@ def check_chest_tabs_index(cls, v: str) -> list[int]:
             raise ValueError("must be a list or a string")
         return sorted([int(x) - 1 for x in v])
 
+    @field_validator("max_stash_tabs")
+    def check_max_stash_tabs(cls, v: str):
+        if v != "6" and v != "7":
```
chrisoro: strange way to check? in check with tuple is much faster and more readable

src/config/models.py
```diff
@@ -273,6 +295,10 @@ class GeneralModel(_IniBaseModel):
         default=True,
         description="Whether to favorite matched items or not",
     )
+    max_stash_tabs: str = Field(
```
chrisoro: why not int? you can then also remove all the int casts

src/tts.py
```diff
@@ -31,9 +32,14 @@ def __init__(self):
 
     def find_item(self) -> None:
         local_cache = []
+        max_stash_tabs = IniConfigLoader().general.max_stash_tabs
         while True:
             data = fix_data(_DATA_QUEUE.get())
             local_cache.append(data)
+            if max_stash_tabs != "7" and data == "Chest 7":
```
chrisoro: Imo I would remove this. The code in here should be as fast as possible and what the user configures is what they get

src/ui/stash.py
```diff
@@ -21,11 +22,14 @@ def __init__(self):
 
     @staticmethod
     def switch_to_tab(tab_idx) -> bool:
-        number_tabs = 7
+        number_tabs = int(IniConfigLoader().general.max_stash_tabs)
         LOGGER.info(f"Switch Stash Tab to: {tab_idx}")
         if tab_idx > (number_tabs - 1):
             return False
-        x, y, w, h = ResManager().roi.tab_slots
+        if number_tabs == 6:
```
chrisoro: here you can see what I mean by bad design. The code does the same, regardless if it is 100 or 2 tabs.  It should be handled in the config/data part so that you don't need conditions anywhere else

src/tts.py
cjshrader: I thought you might leave a comment like this because I had similar thoughts. I do things like this to avoid a million "the stash clicker is clicking incorrectly" comments/issues as nothing tells the user they need to make a change. They don't read the readme.

However I will move this to a subscriber which is the correct way to manage this, I was just trying to get something out as quick as possible because we can't make any releases until this is merged.

src/config/data.py
cjshrader: I don't disagree, but POSITIONS is a giant constant and I was trying to maintain that. Can you explain more how this would look?

src/config/models.py
cjshrader: It originally was an int and looking over the code now the reason I converted it to a string doesn't even matter, I will convert

cjshrader:
I made a few of the changes you recommended, I will make more this evening when I get back from work

src/config/data.py
chrisoro: https://github.com/d4lfteam/d4lf/pull/588/files#diff-03ddc0f629f2a9081a62cc037ff902ca53292131ae160f98efc677e6379e5033R70
and
https://github.com/d4lfteam/d4lf/pull/588/files#diff-03ddc0f629f2a9081a62cc037ff902ca53292131ae160f98efc677e6379e5033R30

Maybe there is a smarter way

chrisoro:
I merged main, resolved the conflicts and deleted some unused mouse code. Only thing left is the stuff in tts.py.

Please change that and then we are a go :)

Issue #578, opened 2025-12-14
unreconized item: galvanic azurite


atheri:
error in the log
```
2025-12-13 | 15:54:43.660 | Thread-111 (_wrapper_run_loot_interaction_method) | ERROR | src.scripts.loot_filter_tts:56 | Error in TTS read_descr. src.tts.LAST_ITEM=['GALVANIC\xa0AZURITE', 'Unique Ring', '750 Item Power', '104 All Resist (-13.1% Toughness)', '+6.0% Critical Strike Chance [5.2 - 6.0]%', '+58.0% Lightning Damage [52.0 - 70.0]%', '8.4% Cooldown Reduction [8.1 - 8.8]%', '+3 to Convulsions [2 - 3]', 'Lightning damage leaves enemies magnetized for 4 seconds causing them to emit Crackling Energy, and increasing all Lightning damage they take from you by 58.0%[x] [40.0 - 60.0]%. If two magnetized enemies hit each other with Crackling Energy they pull each other together.', 'This gem acts as a conduit for aetheric filaments, enabling the wielder to summon atmospheric discharge at will. - Excerpt from Volatus Arcanum: Of Skybinding and Stormcraft', 'Requires Level 60Sorcerer. Only. Unique Equipped', 'Sell Value: 92,439 Gold', 'Mousewheel scroll down', 'Scroll Down', 'Right mouse button']
Traceback (most recent call last):
  File "src\scripts\loot_filter_tts.py", line 51, in check_items
  File "src\item\descr\read_descr_tts.py", line 470, in read_descr
IndexError: Unrecognized unique galvanic azurite. This most likely means the name of it reported from Diablo 4 is wrong. Please report a bug with this message.
```

cjshrader:
Sometimes the TTS engine actually has typos or the wrong name when they tell us the name of the object, which is what's happening here.

I've updated our map of name corrections (This is the second one for galvanic azurite) so it should work now. I don't have a galvanic azurite so can you confirm?

cjshrader:
This is in release 7.6.4, that is

atheri:
I tried with the latest (7.6.5) and it looks to be the same error
```
[...]

PR #590, opened 2025-12-15
(BugFix) another fix for galvanic azurite, using unicode for nbsp

atheri:
another fix for unique 'galvanic azurite'. 

Tested locally on my client with the item, and seems to work.

issue: https://github.com/d4lfteam/d4lf/issues/578


atheri:
interesting that pre-commit changed the unicode `\u00a0` to the actual nbsp character. Makes it a bit less clear imo, but tested it and it still works.

chrisoro (approved):

PR #677, opened 2026-03-31
(New Feature) Addition of Percentage Value to Affixes to loot filter. See readme section "Filtering on percent of affix instead of value" for more information

chrizzocb:


src/config/models.py
```diff
@@ -137,6 +138,21 @@ def name_must_exist(cls, name: str) -> str:
...
+    def percent_validator(cls, v: int) -> int:
+        check_greater_than_zero(v)
+        if v > 100:
+            msg = "must be less than or equal to 100"
+            raise ValueError(msg)
+        return v
+
+    @model_validator(mode="after")
+    def value_and_percent_are_mutually_exclusive(self) -> AffixFilterModel:
+        if self.value is not None and self.minPercentOfAffix:
```
cjshrader:
```suggestion
        if self.value and self.minPercentOfAffix:
```

src/item/descr/read_descr_tts.py
```diff
@@ -126,26 +126,33 @@ def _add_affixes_from_tts_mixed(
     inherent_num, affixes_num = _get_affix_counts(tts_section, item, starting_index)
     affixes = _get_affixes_from_tts_section(tts_section, starting_index, inherent_num + affixes_num)
     aspect_text = _get_aspect_from_tts_section(tts_section, item, starting_index, len(affixes))
+    allow_missing_bloodied_bullet = (
+        item.seasonal_attribute == SeasonalAttribute.bloodied and len(affixes) == len(affix_bullets) + 1
+    )
 
-    # With advanced item compare on we'll actually find more bullets than we need, so we don't rely on them for number of affixes
-    if len(affixes) - 1 > len(affix_bullets):
+    # With advanced item compare on we can find more bullets than affixes, but we still need
```
cjshrader: What problem are you trying to address with this? I just recently reworked this code, I'm not aware of any changes that need to be made in read_descr_tts at all

src/item/descr/texture.py
```diff
@@ -80,6 +80,11 @@ def _find_bullets(
     return sorted(filtered_matches, key=lambda match: match.center[1])
 
 
+def _filter_defined_templates(template_list: list[str]) -> list[str]:
```
cjshrader: Once again what problem are you solving here?

src/item/filter.py
```diff
@@ -425,7 +425,7 @@ def _match_greater_affix_count(expected_min_count: int, item_affixes: list[Affix
         return expected_min_count <= len([x for x in item_affixes if x.type == AffixType.greater])
 
     @staticmethod
-    def _match_aspect_is_in_percent_range(expected_percent: int, item_aspect: Aspect) -> bool:
+    def _match_item_roll_is_in_percent_range(expected_percent: int, item_aspect: Aspect | Affix) -> bool:
```
cjshrader: Please rename item_aspect to item_aspect_or_affix to be more clear

src/item/filter.py
```diff
@@ -459,6 +459,14 @@ def _match_item_aspect_or_affix(
                 expected_aspect.comparison == ComparisonType.smaller and item_aspect.value > expected_aspect.value
             ):
                 return False
+        expected_affix_percent = getattr(expected_aspect, "minPercentOfAffix", 0)
+        if expected_affix_percent:
+            if isinstance(item_aspect, Affix) and item_aspect.type == AffixType.greater:
+                return True
+            if not Filter._match_item_roll_is_in_percent_range(
```
cjshrader: No reason to call Filter.* here, we're already in Filter. Just call _match_item_roll* directly.

tests/item/filter/filter_test.py
```diff
@@ -39,6 +52,40 @@ def test_affixes(_name: str, result: list[str], item: Item, mocker: MockerFixtur
     assert natsorted([match.profile for match in test_filter.should_keep(item).matched]) == natsorted(result)
 
 
+def test_affix_percent_matches_normal_affixes(mocker: MockerFixture):
+    test_filter = _create_mocked_filter(mocker)
```
cjshrader: These tests need to be consistent with the other tests. Filters are in the filters.py file. Items are in the affixes.py file. I know it's confusing but that's how it is.

I suspect this and the one to test greater affixes can be merged

cjshrader (changes requested):

src/gui/profile_editor/affixes_tab.py
```diff
@@ -290,33 +294,43 @@ def sync_min_greater_from_checkboxes(self):
             count = self.count_want_greater_affixes()
             self.min_greater.setValue(count)
 
+    def _ensure_pool_widgets_initialized(self):
```
cjshrader, line 297: Can you give more detail on what this method and the method after it is doing? I don't understand why these are needed to add a new field

cjshrader: Actually I think I'm following

src/gui/profile_editor/affixes_tab.py
```diff
@@ -290,33 +294,43 @@ def sync_min_greater_from_checkboxes(self):
...
+            was_visible = container.contentWidget.isVisible()
+            if container.header.first_expansion:
+                container.expand()
+                if not was_visible:
+                    container.collapse()
+
+    def iter_affix_widgets(self):
+        self._ensure_pool_widgets_initialized()
+
+        for layout_widget in (self.affix_pool_layout, self.inherent_pool_layout):
```
cjshrader: Do we want to do inherents here? I know the original code did that but it seems like an error. Can an inherent be a GA? Would an inherent ever need a minPercentOfAffix?

src/item/descr/read_descr_tts.py
chrizzocb: [src/item/descr/read_descr_tts.py](https://file+.vscode-resource.vscode-cdn.net/c%3A/Users/Christopher/.vscode/extensions/openai.chatgpt-26.325.31654-win32-x64/webview/): removed the added bloodied-specific fallback and restored the prior _add_affixes_from_tts_mixed() mismatch and bullet-assignment logic.
[tests/item/read_descr_season_11_mixed_test.py](https://file+.vscode-resource.vscode-cdn.net/c%3A/Users/Christopher/.vscode/extensions/openai.chatgpt-26.325.31654-win32-x64/webview/): removed the tests that were only added to lock in that new read_descr_tts behavior.

src/item/descr/texture.py
chrizzocb: [src/item/descr/texture.py](https://file+.vscode-resource.vscode-cdn.net/c%3A/Users/Christopher/.vscode/extensions/openai.chatgpt-26.325.31654-win32-x64/webview/): removed _filter_defined_templates() and restored the original template list flow for find_affix_bullets() and find_aspect_bullet().
[tests/item/descr/texture_test.py](https://file+.vscode-resource.vscode-cdn.net/c%3A/Users/Christopher/.vscode/extensions/openai.chatgpt-26.325.31654-win32-x64/webview/): removed the test that existed only to enforce that helper behavior.

src/item/filter.py
chrizzocb: I renamed it to item_aspect_or_affix

src/item/filter.py
chrizzocb: removed @staticmethod from _match_item_aspect_or_affix()
added self to that method
changed the internal call from Filter._match_item_roll_is_in_percent_range(...) to self._match_item_roll_is_in_percent_range(...)

tests/item/filter/filter_test.py
chrizzocb: everything is merged now

src/gui/profile_editor/affixes_tab.py
chrizzocb: the Greater Affix auto-sync label/count, and the bulk Set All Min % / Convert All To Min % actions no longer traverse inherent pools.

chrizzocb: i also deleted the Set All Min % functions

src/item/descr/read_descr_tts.py
cjshrader: You don't always need to remove the code I just need to know what you're trying to do. These changes were just completely unrelated to the filtering changes.

If you were trying to fix recent bugs I want to mention that the AI doesn't really fix the source issue, it is just making the warning/error go away. But we want the warning/error so we know there is a problem, if we remove that it's masking an underlying issue.

src/config/models.py
```diff
@@ -137,6 +138,21 @@ def name_must_exist(cls, name: str) -> str:
             raise ValueError(msg)
         return name
 
+    @field_validator("minPercentOfAffix")
+    def percent_validator(cls, v: int) -> int:
+        check_greater_than_zero(v)
+        if v > 100:
```
cjshrader: Wouldn't this allow values like -1? We should probably require anything from 1 to 100 (as in don't allow 0 either)

src/config/models.py
```diff
@@ -442,7 +458,7 @@ def check_profiles_is_list(cls, v: str) -> list[str]:
         elif not isinstance(v, list):
             msg = "must be a list or a string"
             raise ValueError(msg)
-        return [v.strip() for v in v]
+        return [profile_name for profile_name in (item.strip() for item in v) if profile_name]
```
cjshrader, line 460: This should work, thank you

src/gui/profile_editor/affixes_tab.py
```diff
@@ -329,6 +343,10 @@ def update_greater_count_label(self):
         else:
             self.greater_count_label.setText(f"({count} greater affixes marked)")
 
+    def convert_all_to_min_percent_of_affix(self, percent: int):
```
cjshrader, line 346: If you removed the buttons, is this method still used?

cjshrader: Also I'm fine keeping the buttons if you want the buttons, you're the developer here. Me asking a question doesn't mean a change needs to be made, I'm open to discussion.

src/config/models.py
cjshrader: Actually we should allow 0 because that's our default. But we don't want to allow negative values still.

cjshrader (changes requested):

src/gui/profile_editor/affixes_tab.py
chrizzocb: x

i deleted the other button. we will use convert_all_to_minpercent, for now

src/config/models.py
chrizzocb: changed it.

src/item/descr/read_descr_tts.py
chrizzocb: I was trying to handle a recent bloodied mixed-read case where TTS returns one more affix than find_affix_bullets() finds, which causes _add_affixes_from_tts_mixed() to fail on the last affix. The intent was only to cover that specific mismatch, not to change read_descr_tts more broadly.

That said, I agree this is unrelated to the filtering changes. If you want to keep the current warning/error behavior until we have a concrete reproducer and fix the actual source issue, I’m fine dropping it from this PR.

cjshrader: Ok let's remove it for now. I would like to see a screenshot of this issue and resolve it in a different PR.

cjshrader (changes requested):

src/config/models.py
```diff
@@ -137,6 +138,23 @@ def name_must_exist(cls, name: str) -> str:
             raise ValueError(msg)
         return name
 
+    @field_validator("minPercentOfAffix")
+    def percent_validator(cls, v: int) -> int:
+        if v < 0:
```
cjshrader: Let's make it one check to be more consistent with other checks, like font_size_in_range(...) is for example

cjshrader (changes requested):

src/config/models.py
chrizzocb: [minPercentOfAffix validation in models.py](https://file+.vscode-resource.vscode-cdn.net/c%3A/Users/Christopher/.vscode/extensions/openai.chatgpt-26.325.31654-win32-x64/webview/) written in the same style as the other range validators: one explicit inclusive range check instead of two separate comparisons. The valid range stays the same: 0..100.

src/item/descr/read_descr_tts.py
chrizzocb: look at update 6

cjshrader (commented):
Everything looks good, thank you as always. I just need to give it a test which I hope to do this evening or tomorrow at the latest.

cjshrader (approved):

PR #699, opened 2026-05-10
(S13) Reworked uniques to have better support for variable affixes and renamed Uniques filter to GlobalUniques to be more clear. This is a breaking change, your old filters will need to be recreated. See Affix and GlobalUnique documentation for more information.

cjshrader:


cjshrader:
Merging is on hold until importers and profile editors are finished too

PR #701, opened 2026-05-11
(S13) Small fixes for Mobalytics import and the Warlock class

chrizzocb:


chrizzocb:
paragon naming fix for warlock on mobalytics

src/gui/importer/paragon_export.py
cjshrader, line 389: We maintain a list of classes in gui_common. Do you think we could loop through that list here so that we don't need to maintain a list of every class in two spots?

cjshrader (changes requested):

src/gui/importer/paragon_export.py
cjshrader: You also have a list of classes here as well: https://github.com/d4lfteam/d4lf/blob/07878a5c5ffa6e39bd8f62ca7b241390d1feeb57/src/paragon_overlay.py#L1367

Let's try to get everything to just be the gui_common list to make it easier to maintain. If that list needs to be a map instead that's fine.

cjshrader (changes requested):

cjshrader (approved):
Thanks, that should be a lot easier to maintain

PR #707, opened 2026-05-13
(S13) Removed ComparisonType from all filters. If we need to know if an affix/aspect value should be bigger or smaller we determine it ourselves 

chrizzocb:
ComparisonType duplicated roll-direction data already available from item min_value/max_value, but direct value matching still depended on the persisted/profile-editor comparison field.

Changed files:

src/config/profile_models.py: removed ComparisonType and comparison; list shorthand now accepts only name/value.

src/item/filter.py: direct value thresholds now infer smaller/larger from roll range, matching percent logic.

src/gui/profile_editor/affixes_tab.py: removed comparison controls and model writes.

tests/item/filter/data/affixes.py, filters.py, uniques.py: updated fixtures and added smaller-is-better aspect coverage.

README.md: removed comparison docs and documented automatic direction.

### DIFF
--- a/src/item/filter.py
+++ b/src/item/filter.py
@@ -429,11 +429,13 @@ class Filter:
     def _match_item_aspect_or_affix(
         self,
         expected_aspect: AffixAspectFilterModel | None,
-        item_aspect: Aspect | Affix,
+        item_aspect: Aspect | Affix | None,
         is_fixed_aspect_value: bool = False,
     ) -> bool:
         if expected_aspect is None:
             return True
+        if item_aspect is None:
+            return False
         if expected_aspect.name != item_aspect.name:
             return False
 
--- a/src/item/filter.py
+++ b/src/item/filter.py
@@ -15,7 +15,6 @@ from src.config.profile_models import (
     AffixAspectFilterModel,
     AffixFilterCountModel,
     AffixFilterModel,
-    ComparisonType,
     DynamicItemFilterModel,
     GlobalUniqueModel,
     ProfileModel,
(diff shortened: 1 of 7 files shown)


tests/item/filter/data/uniques.py
```diff
@@ -55,6 +55,7 @@ def __init__(self, rarity=ItemRarity.Unique, item_type: ItemType = ItemType.Shie
 
 uniques_with_affixes = [
     ("matches nothing", [], TestUnique(item_type=ItemType.Amulet, aspect=Aspect(name="dolmen_stone"))),
+    ("unique aspect missing", [], TestUnique(item_type=ItemType.Helm, affixes=[Affix(name="maximum_life", value=641)])),
```
cjshrader: Why would there be a unique with no aspect? That doesn't seem possible. Were you trying to test a rare? Rares are already being tested elsewhere so I'm unclear what this is testing

src/item/filter.py
```diff
@@ -410,14 +412,30 @@ def _match_item_roll_is_in_percent_range(expected_percent: int, item_aspect_or_a
...
     def _match_item_aspect_or_affix(
         self,
         expected_aspect: AffixAspectFilterModel | None,
-        item_aspect: Aspect | Affix,
+        item_aspect: Aspect | Affix | None,
         is_fixed_aspect_value: bool = False,
     ) -> bool:
         if expected_aspect is None:
             return True
+        if item_aspect is None:
```
cjshrader, line 437: I saw you mention/delete that rares were broken, but they were working fine before this change so I'm not clear if this solved anything. I'd like to know what the errors were

cjshrader (changes requested):

src/item/filter.py
chrizzocb: Fair point. I should not have described this as rares being broadly broken. The actual error was an AttributeError: 'NoneType' object has no attribute 'name' from _match_item_aspect_or_affix() when _check_affixes() evaluated a profile rule with uniqueAspect and the current item had item.aspect is None. The logs show this through should_keep -> _check_affixes -> _match_item_aspect_or_affix. I updated the test fixture to represent that as a Rare item hitting an Affixes rule with uniqueAspect, instead of implying a real Unique with no aspect.

tests/item/filter/data/uniques.py
chrizzocb: will update this

PR #710, opened 2026-05-14
(BugFix) Attempted fix for issue that would have us read in the loot filter as TTS information

chrizzocb:
The TTS item detector treated the loot-filter UI text SELECT ALL as an item start because it is all uppercase. That caused the loot-filter dialog text to be parsed as an item. Then read_descr_tts saw Item Power Range, tried to extract a number, got None, and crashed with int(None).

Changed files:

src/tts.py: ignores SELECT ALL as a TTS item start.
src/item/descr/read_descr_tts.py: safely returns None when an Item Power line has no number.
tests/item/read_descr_tts_test.py: adds regression tests for this loot-filter TTS case.


cjshrader (approved):
I could not reproduce the original problem in-game at all no matter what I tried even without these changes, but these changes won't hurt so let's just go with them

PR #711, opened 2026-05-14
(Development) Profiles are sorted alphabetically and AspectUpgrades are shown in block format

chrizzocb:
AspectUpgrades kept source order and YAML dumped scalar lists inline. Imported unique filters were sorted by slot/filter name, not by the unique aspect name.


cjshrader (approved):
This looks good, thank you

Issue #713, opened 2026-05-15
Profile editor will not let you select multiple items types for Affixes


cjshrader:
<img width="3093" height="1654" alt="Image" src="https://github.com/user-attachments/assets/e4bbc23c-770b-46b5-a546-96b4672d8f0e" />

Affixes allows multiple item types for a field, but the profile editor will only let you select one.

It could also be beneficial to add an "All" button that would add all item types to an Affix. This will become useful for keeping ancestral commons.

PR #717, opened 2026-05-16
(BugFix) (Profile Editor) Added ability to select more than one item type for Affixes

chrizzocb:
Root cause: Affix filters already support multiple itemType values, but the profile editor used a single-select combo and overwrote the list with one selected type.

Code change: Replaced the item type selector in src/gui/profile_editor/affixes_tab.py with a checkable dropdown, added All and Reset buttons, and kept updates flowing into the existing config.itemType list.


cjshrader:
This is very good and definitely works. Seeing it in person gave me some ideas to make it more useable for people, I hope you don't mind.

First, I would like it if the user could see all or most of the selections on the main edit profile screen. The best way to do that is in the edit profile screen Item Type is displayed the same way we display profiles in the settings screen: A text box with the selections populated, and then a button with ... that opens a different window. 

image

In that other window, what do you think about separating the items by weapon and nonweapon? In item_type.py is what you'd need to do that, there are methods that say if an item type is a weapon, armor, or jewelry. I'm envisioning two scrollboxes with a list of checkboxes the user can check. We technically consider a shield to be armor, which I think is for evaluation reasons, so we might want to "hardcode" that to be a weapon. Or create a separate offhand list? In which case we'd need to add an "is_offhand" method to item_type.py.

Finally, and this is my fault, I totally forgot that selecting no item types is the same as selecting all item types. We don't need an "All" button. Maybe just a clear button on the popup window and a small note on that window saying if no item types are selected, all item types will be evaluated for this filter.

src/gui/profile_editor/affixes_tab.py
```diff
@@ -51,6 +52,83 @@
 AFFIXES_TABNAME = "Affixes"
 AFFIX_VALUE_MODE = "Value"
 AFFIX_PERCENT_MODE = "Min %"
+ALL_ITEM_TYPES_TEXT = "All item types"
+ITEM_TYPE_HELP_TEXT = "If no item types are selected, all item types will be evaluated for this filter."
+
```
cjshrader: If a constant is only being used once in once place, there's no reason to generate the constant. Let's just have these strings live wherever they are used

src/gui/profile_editor/affixes_tab.py
```diff
@@ -51,6 +52,83 @@
...
+
+def _item_type_display_text(item_type: ItemType) -> str:
+    return item_type.value
```
cjshrader: No use to make a method that only does one thing

src/gui/profile_editor/affixes_tab.py
```diff
@@ -51,6 +52,83 @@
...
+
+def _item_type_summary(item_types: list[ItemType]) -> str:
+    if not item_types:
+        return ALL_ITEM_TYPES_TEXT
+    return ", ".join(_item_type_display_text(item_type) for item_type in item_types)
+
+
+def _is_weapon_picker_item_type(item_type: ItemType) -> bool:
+    return is_weapon(item_type) or item_type == ItemType.Shield
+
```
cjshrader, line 61: Same as above, just no need for a method that only does one thing. It lowers readability

cjshrader (changes requested):

cjshrader:
Closes #713

PR #728, opened 2026-05-18
(Profile Editor) Previously opened profile is now remembered, and made it easier to swap between profiles

chrizzocb:
update3

image


cjshrader:
This is good, I just have two thoughts:

image

I think the "Profile loaded" portion of the screen is unnecessary since you're shown what profile is loaded in the box to the right.

Instead of clicking ... and bringing up a profile selector, what do you think of a dropdown? You just click it and select it, removing a popup menu from the equation. It would have an unselectable "label" for active profiles and inactive profiles so you could tell which is which but you could quickly swap. I guess you'd have an entry that's like "---------Active Profiles--------" and "---------Inactive Profiles---------" or something like that. I'll leave that up to you if you like that better than what you have or not, I just like to minimize clicks but it's a question on which way is the better way to display information to the user.

I see in this update it remembers your last selected one and auto-loads it, very cool.

cjshrader (approved):
This is amazing, thank you

PR #730, opened 2026-05-18
new sigils generation

chrizzocb:
Sigil generation was using StringList files as the primary source. Those lists are not always complete, so some sigil data such as AncestralElites could be missed. The generator also still expected a local D4Companion path.

Code change:

Added D4DataSource in src/tools/gen_data.py to read d4data either from GitHub online by default or from an optional local d4data checkout.
Removed the D4Companion input requirement.
Refactored sigil generation into generate_sigils().
Sigil dungeons now start from json/base/meta/World/DGN_*.wrl.json.
Sigil affixes now start from json/base/meta/DungeonAffix/*.dax.json and only include minor, major, and positive.
StringList files are now resolved from the meta files instead of being the source of truth.
Added sigil affix rarity extraction from color tags where available: Common, Magic, Rare, Legendary, Mythic.
Preserved existing affixes.json behavior for now instead of replacing it with incomplete d4data-derived output.


cjshrader:
Did you run gen_data and see if the files were updated? This doesn't seem right at all, we get affix data from d4companion and that got removed here. I'd expect to see changes to affixes.json. But also that's beyond the scope of the ask for this.

I also don't see AncestralElites in the sigils so either gen_data wasn't run or it doesn't work, I'm unclear which.

Also let's tie the rarity to each affix, in a map or whatever. No reason for them to be a standalone map.

If gen_data was not run please run it and commit the changes. That will give me a good idea if the output is correct. I don't think a complete rewrite of gen_data was needed and that's what we got here.

cjshrader:
Yeah looking at it more things really went rogue. In my original message I'm clear that generate_uniques and generate_aspects should be the example. I would expect the only changes in the code to be to remove the block of code that generates sigils and put it in its own singular method, and that method is changed to start from the item files instead of the string files just like the other two. Everything else is more than I asked for.

cjshrader:
I'd really need my previous comment addressed, and I think ancestral_elites is still missing in all this. That was the main reason for this exercise. I checked the files myself and I think the data is there, but if there's some reason why it can't be gotten just let me know.

But importantly all the other changes needs to be undone. They don't seem to be doing anything.

cjshrader:
The rarity is very good though, I have some ideas for that for the future, very glad you could get that

PR #736, opened 2026-05-19
(Dev) Our sigil data should now be more reliable, and added positive sigil rarity for future functionality.

chrizzocb:
DungeonAffix rarity is not in the .dax.json fields; it is encoded in AffixName StringList color tags like {c_white}, {c_magic}, {c_rare}, {c_legendary}, {c_mythic}.

gen_data.py built sigils from StringList globs first. I moved sigil generation into generate_sigils() and now start from d4data/json/base/meta/World and d4data/json/base/meta/DungeonAffix, then resolve the matching StringList files. CoreTOC.dat.json is used as a fallback name mapping.

tested via
uv run python -m src.tools.gen_data d4data D4Companion
Gen Aspects for enUS
Gen Uniques for enUS
Gen Sigils for enUS
Gen Tributes for enUS
Gen ItemTypes for enUS
Gen Tooltips for enUS
Gen Affixes for enUS


src/tools/gen_data.py
```diff
@@ -309,6 +265,135 @@ def generate_aspects(d4data_dir, language):
         json_file.write("\n")
 
 
+def generate_sigils(d4data_dir, language):
+    print(f"Gen Sigils for {language}")
+    sigil_dict = {"dungeons": {}, "minor": {}, "major": {}, "positive": {}}
+    sigil_rarity_dict = {"minor": {}, "major": {}, "positive": {}}
+    core_toc_names = load_core_toc_names(d4data_dir)
```
cjshrader: Let's remove loading of the core_toc file, that's a 40 mb file. I only mentioned it as a way for Codex to figure out how to find things (it's the same process I use) but didn't mean for it to be actually used in the code.

src/tools/gen_data.py
```diff
@@ -309,6 +265,135 @@ def generate_aspects(d4data_dir, language):
...
+    json_files = sorted(d4data_dir.glob(pattern, case_sensitive=False))
+    for json_file in json_files:
+        affix_type = json_file.stem.split("_", maxsplit=1)[0].lower().strip()
+        if affix_type not in sigil_dict or affix_type == "dungeons":
+            continue
+
+        with Path(json_file).open(encoding="utf-8") as file:
+            affix_data = json.load(file)
+
+        string_list_file = find_string_list_file(
```
cjshrader: I don't think we entirely need this "find_string_list_file".  generate_uniques does this with like 3 lines of code and no submethods.

src/tools/gen_data.py
```diff
@@ -309,6 +265,135 @@ def generate_aspects(d4data_dir, language):
         json_file.write("\n")
 
 
+def generate_sigils(d4data_dir, language):
+    print(f"Gen Sigils for {language}")
+    sigil_dict = {"dungeons": {}, "minor": {}, "major": {}, "positive": {}}
+    sigil_rarity_dict = {"minor": {}, "major": {}, "positive": {}}
```
cjshrader: I think only positive can have a rarity, that's what I'm seeing from what you've found

cjshrader: Actually instead of separating rarity into minor/major/positive at all let's just put them all in the rarities field at the top level. That'll make it easy to see later and check "I have xyz sigil what rarity is it"

cjshrader (changes requested):
Overall this is once again very good, thank you for doing this. This change ensures we won't miss sigils in the future, and the rarity change will allow me (or you if you're feeling frisky) to let people keep all mythic/legendary sigils which is nice.

src/tools/gen_data.py
chrizzocb: see update 2

removed the TOC loader and related source_data plumbing.

src/tools/gen_data.py
chrizzocb: see update2
now builds the dungeon and affix string-list paths inline and removes the helper.

src/tools/gen_data.py
chrizzocb: see update 3

rarities was grouped by sigil category even though the useful lookup is directly from sigil key to rarity.

src/tools/gen_data.py
```diff
@@ -309,6 +265,89 @@ def generate_aspects(d4data_dir, language):
...
+    print(f"Gen Sigils for {language}")
+    sigil_dict = {"dungeons": {}, "minor": {}, "major": {}, "positive": {}}
+    sigil_rarity_dict = {}
+    string_list_dir = d4data_dir / f"json/{language}_Text/meta/StringList"
+
+    pattern = "json/base/meta/World/DGN_*.wrl.json"
+    json_files = sorted(d4data_dir.glob(pattern, case_sensitive=False))
+    for json_file in json_files:
+        string_list_file = string_list_dir / f"World_{json_file.stem}.stl.json"
+        if not string_list_file.exists():
```
cjshrader, line 302: I get the error that it can't find string list for every single dungeon and affix and no sigils are generated. Do you get that error?

cjshrader (changes requested):

src/tools/gen_data.py
chrizzocb: solved in update 5

PR #737, opened 2026-05-19
(New Feature) You can now disable a hotkey if you want it completely turned off.

chrizzocb:
for more information ask schoof on discord, he had problems with those keys

Root cause: Hotkey handling relied on keyboard character names, which can mis-handle German/dead keys like ä, ö, and ^. Arrow keys also reached focused dialog buttons first, so Qt used them for focus navigation instead of recording them as hotkeys.

Code change: Added shared hotkey normalization for validation/registration, allow disabling optional advanced hotkeys, added picker support for special keys, and intercept arrow-key presses on hotkey dialog buttons.

Behavior impact: Users can bind German layout keys, arrow keys, and modifier combos like shift+left, and can clear optional action hotkeys.

### DIFF
--- a/src/config/helper.py
+++ b/src/config/helper.py
@@ -5,6 +5,20 @@ if sys.platform != "darwin":
     import keyboard
 
 
+def _drop_negative_scan_codes(scan_codes: tuple[int, ...]) -> tuple[int, ...]:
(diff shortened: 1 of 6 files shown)


cjshrader (approved):
Gonna be honest I don't really understand this one but it seems to work

PR #739, opened 2026-05-20
(S13/Dev) Generate set data for charms for future charm support

chrizzocb:
getting charm set data out of d4data local source


src/tools/gen_data.py
```diff
@@ -114,6 +130,7 @@ def main(d4data_dir: Path, companion_app_dir: Path):
         file_names = [
             f"assets/lang/{lang}/affixes.json",
             f"assets/lang/{lang}/aspects.json",
+            "assets/sets.json",
```
cjshrader: This should still be in the lang folder, even if we don't support other languages right now

src/tools/gen_data.py
```diff
@@ -105,6 +105,22 @@ def check_ms(input_string) -> str:
     return input_string.replace("{d}", "")
 
 
+def clean_item_name(name: str) -> str:
```
cjshrader, line 108: Ok I give in, it can make the small methods. However I don't like a single-use method. Let's make sure the uniques and aspects are updated to use this method, and anywhere else if needed. This may need to be updated to replace more info. 

If get_string_list_name can be used in multiple places then let's update those too.

cjshrader (changes requested):

src/tools/gen_data.py
chrizzocb: solved in update2

src/tools/gen_data.py
chrizzocb: update2

src/tools/gen_data.py now uses get_string_list_name for aspects, uniques, and sets.

cjshrader:
You just need to run gen_data again and remove the old committed sets file and add the new one. When making changes to gen_data this should always be the process.

There should be *0* changes to the generated uniques and aspects. If there are, you'll need to tweak your cleanup name method as something got lost.

cjshrader (approved):
It's beautiful, thank you

PR #741, opened 2026-05-21
Affix from d4data

chrizzocb:


assets/lang/enUS/affixes.json
```diff
@@ -1,25 +1,23 @@
 {
+    "abyss_cooldown_reduction": "abyss cooldown reduction",
     "abyss_damage": "abyss damage",
-    "advance_resource_generation": "advance resource generation",
```
cjshrader, line 3: Reminder about this comment from discord, because it's very important:

----
So there's something important here for affixes. We can't break existing builds, we need to build the affix strings the exact same way diablo 4 companion does. I assume this code does something but I don't know .cs, codex might need to find it: https://github.com/josdemmers/Diablo4Companion/blob/master/D4Companion.Services/AffixManager.cs

So we want to do it in our "style," (Starting at the json/base/meta files), but the affix string names in the file should not change so we need to format them the same way that D4Companion does
----

You have d4companion checked out locally now, that's probably the easiest way for codex to search it and figure out how they export affixes. We need the formatting to be the same.

Some examples I see:
berserking_duration became berserking_duration_duration
bone_spirit_damage became tempered_damage_skill_bone_spirit_tier_damage I think
chance_for_rapid_fire_projectiles_to_cast_twice became chance_for_talisman_charm_affix_unique_projectiles_skill_rapid_fire_projectiles_to_cast_twice
damage_to_immobilized_enemies became damage_to_to_immobilized_enemies

This is probably the most important thing. If we can't format them like d4companion does then we probably can't do this feature. But I think we have access to all their code so I think it is possible.

If you are seeing a lot of changes in this file (100s of lines, like here) then that means it isn't working.

src/tools/gen_data.py
```diff
@@ -35,6 +35,9 @@
     "Wand",
 ]
 
+DAMAGE_TYPES = ["Cold", "Fire", "Holy", "Lightning", "NonPhysical", "Physical", "Poison", "Shadow"]
```
cjshrader: I really don't like maintaining these lists, is there a way to generate them dynamically from d4data? Especially the resource types but preferably all of it.

src/tools/gen_data.py
```diff
@@ -226,37 +229,239 @@ def main(d4data_dir: Path, companion_app_dir: Path):
...
-                else:
-                    affix_dict[key] = value
-        with Path(D4LF_BASE_DIR / f"assets/lang/{language}/affixes.json").open("w", encoding="utf-8") as json_file:
-            json.dump(affix_dict, json_file, indent=4, ensure_ascii=False, sort_keys=True)
-            json_file.write("\n")
-
+        if companion_app_dir is None:
+            generate_affixes(d4data_dir, language)
+        else:
+            generate_affixes_from_companion(companion_app_dir, language)
```
cjshrader: The goal is to completely remove generating affixes from companion, so this won't be needed. You may want to leave this in the short term though so that you can generate from both and see the differences. That's up to you

src/tools/gen_data.py
```diff
@@ -226,37 +229,239 @@ def main(d4data_dir: Path, companion_app_dir: Path):
...
+        description = _build_affix_description(
+            affix_name=affix_name,
+            affix_data=affix_data,
+            attribute_descriptions=attribute_descriptions,
+            power_names={},
+            skill_tag_names=skill_tag_names,
+        )
+        if description is None:
+            continue
+        name = description.replace(",", "").replace(" ", "_")
```
cjshrader: Shouldn't this use the clean_item_name method?

cjshrader (changes requested):

assets/lang/enUS/affixes.json
cjshrader: By the way this is why the tests fail, the affixes changed. Let me know if you need more information.

src/tools/gen_data.py
chrizzocb: see update2

src/tools/gen_data.py
chrizzocb: see update3

src/tools/gen_data.py
chrizzocb: see update4

assets/lang/enUS/affixes.json
chrizzocb: some changes in update5

PR #753, opened 2026-05-25
(Profile Editor) Now supports multiple uniqueAspects

chrizzocb:
-Updated affixes_tab so uniqueAspect is edited as a list instead of a single object.
-Added add/remove row controls for multiple unique aspects, with the same value/min-percent threshold behavior as before.
-Moved Unique Aspects into a collapsible container like Affix Pool, opened by default.
-Fixed a stray maxroll.py import from gui... to src.gui... so python -m src.main no longer fails on that import.

### DIFF
--- a/src/gui/importer/maxroll.py
+++ b/src/gui/importer/maxroll.py
@@ -5,7 +5,6 @@ import re
 import lxml.html
 
 import src.logger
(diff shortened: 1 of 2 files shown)


cjshrader (approved):
Thank you so much

PR #755, opened 2026-05-25
(Dev) Affix data is now imported directly from D4Data instead of relying on Diablo 4 Companion

chrizzocb:
validated with uv run python -m src.tools.gen_data d4data

gen_data.py still depended on D4Companion/Data/Affixes.*.json; AffixManager.cs only consumes that data, while the real string-building logic is in D4DataParser/Parsers/AffixParser.cs.

Code change: Ported the relevant Companion affix parsing/localisation logic into src/tools/gen_data.py, starting from json/base/meta/Affix/*.json, while keeping the generated assets/lang/enUS/affixes.json format unchanged. Added a few existing fallback keys to custom_affixes_enUS.json.

Impact: Existing affix keys/values are preserved. The generated enUS affix output matches the current file exactly.


src/tools/gen_data.py
```diff
@@ -128,6 +360,295 @@ def clean_item_name(name: str) -> str:
     return check_ms(clean_name)
 
 
+def load_json_file(json_file: Path):
+    with json_file.open(encoding="utf-8") as file:
+        return json.load(file)
+
+
+def unsigned_int(value: int) -> int:
```
cjshrader: If a method is only one line long let's remove it please, we can just repeat this code where necessary

src/tools/gen_data.py
```diff
@@ -43,6 +43,238 @@
     "c_mythic": "Mythic",
 }
 
+CROWD_CONTROL_TYPES = {
```
cjshrader: I knew a lot of these maps would be necessary and expected them, but do you think we could move them to a different file at least? Maybe a gen_data_helpers.py?

cjshrader (changes requested):
(Removed, this was addressed)

josdemmers:
Noticed this PR when looking through the repo.
The Holy Damage affixes was fixed in the data parser earlier this month. Previously I localized it as non-physical.
The mapping I used to find the correct localization strings for damage types was no longer correct since season 13.

Here is an example for Holy Damage ingame:
image

cjshrader:
Thank you @josdemmers! So it is a real affix, it's just a temper. One struggle I've had is I can't find an authoritative list of every affix in the game to compare against our list.

I don't know if you know this but we used to just get our affixes from your data. Which has worked fine of course, but I didn't like having a reliance on two different repos (yours and d4data) for ours to function. I asked @chrizzocb here to unleash his AI to break that dependency, but in a way I also wanted to keep it how it was because it meant you had to handle missing affixes instead of me. Oh well.

josdemmers:
In the past there was this list: https://www.d4craft.com/affixes
Too bad the last update was from season 5. I'm not aware of a similar website that is up-to-date.

But feel free to use my parsed data when needed. That's why it's open source.

Like you, I also wanted to look into the affixes/aspects to check if they are actually being used in game. But haven't found a good solution for that yet.
Season 13 (Expansion) especially took some time because both pre- and current season aspects remain in the game data. So you now have aspects with the exact same name but with different descriptions.

src/tools/gen_data.py
chrizzocb: update2

unsigned_int() was a one-line wrapper around value % (2**32) and only used inside src/tools/gen_data.py

src/tools/gen_data.py
chrizzocb: see update3

src/tools/gen_data.py
```diff
@@ -2,9 +2,48 @@
...
+    item_requirements: dict[str, str]
+    necromancer_army: dict[str, str]
+    power_by_sno: dict[int, str]
+    skill_tags: dict[str, str]
+    skill_tags_by_sno: dict[int, list[str]]
+    ui_tooltips: dict[str, str]
+    weapon_types_by_sno: dict[int, str]
+
+
 GEAR_TYPES = [
```
cjshrader: Let's move this to gen_data_helpers too, no reason not to.

src/tools/gen_data.py
```diff
@@ -128,6 +159,296 @@ def clean_item_name(name: str) -> str:
...
+    power_file_name = power_by_sno.get(sno, "")
+    return Path(power_file_name).stem
+
+
+def get_first_gbid_name(gbid_by_sno: dict[int, list[str]], sno: int) -> str:
+    names = gbid_by_sno.get(sno, [])
+    return names[0] if names else ""
+
+
+def dot_damage_type_id(sno: int) -> str:
```
cjshrader: I think I wasn't clear enough earlier, please ask Codex:

If a method has only 1 line, just remove it.

If a method has fewer than 5 lines and is used in only one place, also remove it and just put all that code in-line.

src/tools/gen_data_helpers.py
```diff
@@ -0,0 +1,239 @@
+SIGIL_RARITY_COLOR_TAGS = {
```
cjshrader, line 1: I don't look forward to maintaining these but it's much cleaner to put them here, thank you.

cjshrader (changes requested):

src/tools/gen_data.py
chrizzocb: update4

src/tools/gen_data.py
chrizzocb: update5
src/tools/gen_data.py still had short helpers matching the maintainer rule: localise, dot_damage_type_id, replace_formula_values, and extract_sigil_rarity. I inlined those. Short helpers used in multiple places were left intact.

cjshrader (approved):

PR #762, opened 2026-05-29
(Mobalytics) Fixed Mobalytics import yet again. Expect a lot of whack-a-mole with them for the future.

chrizzocb:
Root cause:
Mobalytics now serves the importer through a Cloudflare challenge in normal headless Selenium, so the importer never reached the build data script.

Code change:
Switched Mobalytics to the existing UC/SeleniumBase path, fixed that path to return a real Driver, made preloaded-state parsing more tolerant, and corrected ProfileModel assignment from AspectUpgrades alias to aspect_upgrades.

Impact:
Mobalytics imports can pass Cloudflare and complete profile generation again. D4Builds/Maxroll behavior is unchanged; the shared UC fix may also help Diablo Trade.

### DIFF
--- a/src/gui/importer/mobalytics.py
+++ b/src/gui/importer/mobalytics.py
@@ -295,7 +295,7 @@ def _extract_mobalytics_preloaded_state(script_text: str) -> dict | None:
 def _log_mobalytics_page_diagnostics(driver: ChromiumDriver, page_source: str, script_count: int) -> None:
     page_source_casefold = page_source.casefold()
     matched_markers = [marker for marker in PAGE_DIAGNOSTIC_MARKERS if marker.casefold() in page_source_casefold]
-    LOGGER.error(
+    LOGGER.debug(
         "Mobalytics page diagnostics: current_url=%r title=%r page_source_length=%s script_count=%s markers=%s",
         _read_mobalytics_driver_value(driver, "current_url"),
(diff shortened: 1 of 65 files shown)


src/gui/importer/mobalytics.py
```diff
@@ -256,6 +273,46 @@ def _fix_input_url(url: str) -> str:
...
+    except json.JSONDecodeError:
+        return None
+    return data if isinstance(data, dict) else None
+
+
+def _log_mobalytics_page_diagnostics(driver: ChromiumDriver, page_source: str, script_count: int) -> None:
+    page_source_casefold = page_source.casefold()
+    matched_markers = [marker for marker in PAGE_DIAGNOSTIC_MARKERS if marker.casefold() in page_source_casefold]
+    LOGGER.error(
+        "Mobalytics page diagnostics: current_url=%r title=%r page_source_length=%s script_count=%s markers=%s",
```
cjshrader, line 299: This is visible to the user and needs to be a lot less technical. Or needs to be logged to debug

chrizzocb: [Update mobalytics.py] change = log to debug
Diagnostic page details were logged with LOGGER.error, making them user-visible instead of debug-only

cjshrader (approved):

cjshrader:
Thank you for once again fixing mobalytics, what a pain it is

PR #775, opened 2026-06-01
mobalytics test

chrizzocb:
added a fallback method for mobalytics import


cjshrader:
This worked for me, but can I get more explanation on why it works? I'm worried it'll break again in a week.

Also when it's done it gets this error:

2026-06-02 20:54:58,148 | Thread-2 (detect_window) | CRITICAL | src.logger:133 | Unhandled exception caused by thread 'Thread-2 (detect_window)'
Traceback (most recent call last):
  File "C:\Users\cjshr\AppData\Roaming\uv\python\cpython-3.14.0-windows-x86_64-none\Lib\threading.py", line 1081, in _bootstrap_inner
    self._context.run(self.run)
    ~~~~~~~~~~~~~~~~~^^^^^^^^^^
  File "C:\Users\cjshr\AppData\Roaming\uv\python\cpython-3.14.0-windows-x86_64-none\Lib\threading.py", line 1023, in run
    self._target(*self._args, **self._kwargs)
    ~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\cjshr\code\d4lf\src\utils\window.py", line 79, in detect_window
    find_and_set_window_position(window_spec)
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^
  File "C:\Users\cjshr\code\d4lf\src\utils\window.py", line 84, in find_and_set_window_position
    hwnd = get_window_spec_id(window_spec)
  File "C:\Users\cjshr\code\d4lf\src\utils\window.py", line 46, in get_window_spec_id
    if window_spec.match(hwnd):
       ~~~~~~~~~~~~~~~~~^^^^^^
  File "C:\Users\cjshr\code\d4lf\src\utils\window.py", line 35, in match
    return _get_process_from_window_name(hwnd).casefold() == self.process_name.casefold() and window_name_ok
           ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^
  File "C:\Users\cjshr\code\d4lf\src\utils\window.py", line 62, in _get_process_from_window_name
    return psutil.Process(pid).name().lower()
           ~~~~~~~~~~~~~~^^^^^
  File "C:\Users\cjshr\code\d4lf\.venv\Lib\site-packages\psutil\__init__.py", line 314, in __init__
    self._init(pid)
    ~~~~~~~~~~^^^^^
  File "C:\Users\cjshr\code\d4lf\.venv\Lib\site-packages\psutil\__init__.py", line 322, in _init
    raise ValueError(msg)
ValueError: pid must be a positive integer (got -315152480)

chrizzocb:
Why the Mobalytics visible-browser fallback works:

Mobalytics uses server-side rendering — all build data is inlined in the HTML as window.__PRELOADED_STATE__ = {…}. We just need the full page source to extract it; no API calls, no DOM interaction.

The headless browser gets blocked intermittently by Cloudflare anti-bot. When that happens, the fallback opens a real visible browser with three key properties:

UC mode (SeleniumBase undetected-chromedriver) — patches the driver to avoid bot fingerprinting
Non-headless — if Cloudflare still shows a challenge, the user solves it manually (180s timeout)
Persistent browser profile (~/.d4lf/browser/mobalytics/) — Cloudflare cookies survive between imports, so after one manual solve, subsequent imports typically pass without intervention
What would actually break it:

Mobalytics removing __PRELOADED_STATE__ entirely (major architecture change on their side, unlikely to happen quietly)
Mobalytics restructuring the JSON schema inside __PRELOADED_STATE__ (the jsonpath queries would need updating)
What won't break it:

Cloudflare changing challenges → user solves them manually in the visible browser
Mobalytics changing their page layout/CSS → we only read the script tag, not the DOM
Stricter headless detection → the fallback is explicitly non-headless with profile reuse
The loading mechanism is resilient. The fragile part is the downstream JSON parsing, which is separate from this fallback logic.

src/gui/importer/mobalytics.py
```diff
@@ -45,18 +45,18 @@
...
-    "captcha",
-    "cloudflare",
-    "access denied",
-    "forbidden",
-    "just a moment",
+    *BLOCKED_PAGE_MARKERS,
 )
+HEADLESS_WAIT_SECONDS = 10
+VISIBLE_BROWSER_WAIT_SECONDS = 180
+FORCE_VISIBLE_FALLBACK_FOR_TESTING = False
```
chrisoro: **Debug flag in production code.** `FORCE_VISIBLE_FALLBACK_FOR_TESTING` should not exist as a module-level toggle in shipped code. It couples the test workaround tightly to runtime logic and risks accidental misuse.

Consider using dependency injection instead — e.g. accept an optional `_force_visible: bool = False` parameter on `_load_mobalytics_preloaded_state` (or accept a callable for the fallback), so tests can pass `True` without touching module state.

tests/gui/importer/test_gui_common.py
```diff
@@ -1,7 +1,20 @@
 from src.config.profile_models import ProfileModel
+from src.gui.importer import gui_common
 from src.gui.importer.gui_common import _to_yaml_str, build_default_profile_file_name
 
 
+class _UserDataDir:
+    mkdir_called = False
+    mkdir_kwargs = {}
```
chrisoro: **Mutable class-level attributes cause shared state between test instances.** `mkdir_called` and `mkdir_kwargs` are class attributes, not instance attributes. If `_UserDataDir` is instantiated more than once in the same test session, mutations from one instance will leak into others.

Fix by moving them to `__init__`:
```python
class _UserDataDir:
    def __init__(self) -> None:
        self.mkdir_called = False
        self.mkdir_kwargs = {}
    ...
```

src/gui/importer/mobalytics.py
chrizzocb: see update2

src/gui/importer/mobalytics.py: removed the module flag and added keyword-only _force_visible: bool = False injection on _load_mobalytics_preloaded_state.
tests/gui/importer/test_mobalytics.py: added focused coverage for forcing the visible fallback without touching module state.

tests/gui/importer/test_gui_common.py
chrizzocb: see update2
tests/gui/importer/test_gui_common.py: moved those defaults into __init__

cjshrader:
Thanks for making this change. Luckily, mobalytics is going to whitelist us so I don't think we'll need it.

PR #778, opened 2026-06-04
(Paragon Overlay) When the overlay is locked, mouse clicks on the green squares are no longer blocked by the overlay

chrizzocb:
Root Cause: In Tkinter on Windows, winfo_id() returns the window handle (HWND) of the internal widget frame instead of the outermost OS wrapper window. Applying extended style changes (WS_EX_TRANSPARENT) to this child handle has no effect because the parent wrapper still catches mouse clicks. Furthermore, Windows caches the styles of active windows; changes do not apply immediately unless forced.

Code Change: Updated src/paragon_overlay.py to:

Retrieve the outermost top-level wrapper HWND using win32gui.GetAncestor(int(self.winfo_id()), win32con.GA_ROOT).
Dynamically toggle the WS_EX_TRANSPARENT style on this wrapper window in a 100ms polling loop, only when the grid is locked and the mouse cursor is not hovering over the control panel or active popup sub-menus.
Force immediate application of the style changes using win32gui.SetWindowPos with the SWP_FRAMECHANGED flag (preserving position/size/Z-order).
Behavior Impact: When the grid is locked, clicks on the green boxes, lines, and transparent canvas pass directly to the game underneath. Hovering the mouse back over the left control panel immediately restores overlay interactivity. Unlocking the grid instantly disables click-through so the overlay can be repositioned.

### DIFF
--- a/src/__init__.py
+++ b/src/__init__.py
@@ -2,4 +2,4 @@ import concurrent.futures
 
 TP = concurrent.futures.ThreadPoolExecutor()
 
-__version__ = "9.2.5"
+__version__ = "9.2.6"
--- a/src/paragon_overlay.py
+++ b/src/paragon_overlay.py
@@ -632,7 +632,7 @@ class ParagonOverlay(tk.Toplevel):
 
         self.btn_settings = _tk_btn(
             btn_cont,
-            text="ParagonOverlay⚙ ▼",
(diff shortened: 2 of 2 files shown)


cjshrader (approved):
Amazing work and it works perfectly, I didn't think this was possible and it's really made the overlay perfect

PR #785, opened 2026-06-15
(Dev) Implement typed Paragon payload schema and validation

### LINKED ISSUE #784: Paragon needs to be a pydantic model
Right now the way we store the paragon board in a profile is stored as "dict[str, object] | list[dict[str, object]]"

This is not good and I missed it when paragon was implemented. It needs to be its own pydantic model, you can't just put whatever you want in there.

This should be possible by telling whatever LLM to look at where the paragon is used and to give it one from an imported model and tell it to just do it.

chrisoro:
See .scratch/typed-paragon-payload/PRD.md

fixes https://github.com/d4lfteam/d4lf/issues/784

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -4,7 +4,7 @@ default_language_version:
 minimum_prek_version: '0.3.0'
 repos:
   - repo: https://github.com/astral-sh/uv-pre-commit
-    rev: 0.11.16
+    rev: 0.11.21
     hooks:
       - id: uv-lock
         priority: &group1 4294967294
@@ -32,7 +32,7 @@ repos:
         priority: *group1
         files: \.(cpp|h)$
   - repo: https://github.com/pre-commit/pre-commit-hooks
-    rev: v5.0.0
+    rev: v6.0.0
     hooks:
       - id: check-added-large-files
         priority: &read-only 4294967295
@@ -62,7 +62,7 @@ repos:
         priority: *group1
       - id: trailing-whitespace
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.15.15
+    rev: v0.15.17
     hooks:
       - id: ruff-format
         priority: *group1
--- a/src/config/profile_models.py
+++ b/src/config/profile_models.py
@@ -414,7 +414,7 @@ class ParagonPayloadModel(BaseModel):
 
     @model_validator(mode="after")
     def paragon_boards_list_must_not_be_empty(self) -> ParagonPayloadModel:
-        if not self.paragon_boards_list:
+        if not self.paragon_boards_list or any(not step for step in self.paragon_boards_list):
             msg = "ParagonBoardsList must not be empty"
             raise ValueError(msg)
         return self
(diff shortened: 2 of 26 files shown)


src/gui/importer/gui_common.py
```diff
@@ -298,27 +298,39 @@ def add_to_profiles(build_name):
 
 # Built in to_yaml_str does not preserve the order of the attributes of the model, which is important for uniques
 def _to_yaml_str(profile: ProfileModel, exclude_defaults: bool, exclude: set[str]) -> str:
-    str_val = profile.model_dump_json(by_alias=False, exclude_defaults=exclude_defaults, exclude=exclude)
+    str_val = profile.model_dump_json(
+        by_alias=True, exclude_defaults=exclude_defaults, exclude_none=True, exclude=exclude
```
chrisoro: L302: 🟡 risk: `by_alias=True` changes every saved profile key, not just Paragon. Keep existing profile YAML shape and serialize only `Paragon` with aliases.

src/config/profile_models.py
```diff
@@ -315,6 +316,110 @@ def parse_rarities(cls, data: str | list[str]) -> list[str]:
...
+        if all(not isinstance(step, list) for step in boards_list):
+            normalized = dict(data)
+            normalized.pop(key, None)
+            normalized["ParagonBoardsList"] = [boards_list]
+            return normalized
+        return data
+
+    @model_validator(mode="after")
+    def paragon_boards_list_must_not_be_empty(self) -> ParagonPayloadModel:
+        if not self.paragon_boards_list:
```
chrisoro: L417: 🔴 bug: `ParagonBoardsList=[[]]` validates, so D4Builds can save a no-board payload. Reject empty steps or return `[]` before building.

src/config/profile_models.py
```diff
@@ -344,3 +449,34 @@ def aspects_must_exist(self) -> ProfileModel:
...
+        if not isinstance(paragon[0], dict):
+            msg = "Paragon legacy list entries must be objects"
+            raise ValueError(msg)
+        return {**data, key: paragon[0]}
+
+    @field_serializer("paragon", when_used="json-unless-none")
+    def serialize_paragon(self, paragon: ParagonPayloadModel | None) -> object:
+        if paragon is None:
+            return None
+        return paragon.model_dump(mode="python", by_alias=True)
```
chrisoro: L482: 🟡 risk: nested `exclude_none`/`exclude_defaults` are bypassed, so stripped profiles serialize `Source:`, `GeneratedAt:`, and default board fields. Pass those exclusions into `model_dump`.

.pre-commit-config.yaml
```diff
@@ -4,7 +4,7 @@ default_language_version:
 minimum_prek_version: '0.3.0'
 repos:
   - repo: https://github.com/astral-sh/uv-pre-commit
-    rev: 0.11.16
+    rev: 0.11.21
```
chrisoro, line 7: L7: 🔵 nit: pre-commit and `uv.lock` upgrades are unrelated to typed Paragon payloads. Move dependency/tool churn to a separate PR.

cjshrader (approved):
It's good, this is not one of those things that was super important because we're never going to allow people to edit a paragon board in the editor but at the same time having it just be a big ol' dict was just no good.

Issue #502, opened 2025-05-23
config editor for sigils doesn't cope with top level affixes


slippycheeze:
I blacklist a handful of affixes on sigils globally; I just hate `drifting_shade` that much.  This doesn't work with the current sigil editor, which seems focused on "dungeon, then affixes".

cjshrader:
@Shebour, I wager there might be other quirks missing with sigils. We support some complicated things that I never personally understood or used. I do see you have implemented the "conditons" though so maybe you got them.

In this case though I do believe everything in sigils.json needs to be supported as a whitelist/blacklist option.

cjshrader:
Just want to mention we're still going to fix this, this is the last hurdle in the profile editor being completely useable

PR #763, opened 2026-05-29
(Finally) Seals and charms are now supported, both for filtering and in the importers. Profile editor support is coming soon, see the readme for more information. Make suggestions for improvements.

chrizzocb:
ProfileModel rejected Seals and Charms because those sections were not defined. TTS could classify HoradricSeal/Charm, but read_descr treated them as unsupported non-equipment and returned None. Changed files:

src/config/profile_models.py
src/item/data/item_type.py
src/item/descr/read_descr_tts.py
src/item/filter.py
tests/config/models_test.py
tests/item/read_descr_tts_test.py
tests/item/filter/filter_test.py

PR #786, opened 2026-06-16
(New Feature) Add rarity filter for affixes, sigils, and tributes. This also fixes sigils in the profile editor

### LINKED ISSUE #771: Add ability to filter on rarity (Including sigils)
Every object that has a rarity should be able to be filtered on rarity. This functionality was not very relevant before, but with crafting there's a reason to want a common/magic/rare that matches where you would not want a legendary. Rarity would be a list in the model and you can select multiple rarities. An empty list would parse as all rarities.

This should apply to affixes, sigils, and tributes (I believe it already exists for tributes).

It is ok to tackle affixes first in one release, then sigils separately as they are a bit more complex. We do not get sigil rarity directly from the object in D4 itself, but have recently been storing rarities in the currently-unused rarities mapping in sigils.json. If one of the sigil "affixes" is in that map then we can assume that's the rarity of the sigil.

The importers would not need to be updated.

The profile editor for affixes would need to be updated. The one for sigils depends on whether or not it's been fixed by then. Right now it is not working (https://github.com/d4lfteam/d4lf/issues/502) but if that's fixed then the ability to add rarity will need to be added.

Or feel free to fix it first in a separate PR, then come back to this.

chrisoro:
fixes https://github.com/d4lfteam/d4lf/issues/771 https://github.com/d4lfteam/d4lf/issues/502


src/gui/models/dialog.py
```diff
@@ -617,3 +640,47 @@ def __init__(self, parent=None):
...
+        group_box = QGroupBox("Rarities")
+        group_layout = QVBoxLayout(group_box)
+        for rarity in ItemRarity:
+            checkbox = QCheckBox(rarity.value)
+            checkbox.setChecked(rarity in selected_rarity_set)
+            self.checkboxes[rarity] = checkbox
+            group_layout.addWidget(checkbox)
+        layout.addWidget(group_box)
+
+        note_label = QLabel("If no rarities are selected, all rarities will be evaluated for this filter.")
```
cjshrader:
```suggestion
        note_label = QLabel("If no rarities are selected, all rarities will be kept for this filter.")
```

src/gui/profile_editor/sigils_tab.py
```diff
@@ -75,34 +85,36 @@ def setup_ui(self):
         self.sigil_name_combo.setEditable(True)
         self.sigil_name_combo.setInsertPolicy(QComboBox.InsertPolicy.NoInsert)
         self.sigil_name_combo.completer().setCompletionMode(QCompleter.CompletionMode.PopupCompletion)
-        self.sigil_name_combo.addItems(sorted(Dataloader().affix_sigil_dict_all["dungeons"].values()))
+        self.sigil_name_combo.addItems(sorted(sigil_name_dict_for_kind(self.kind).values()))
```
cjshrader, line 88: The dialog for this worked fine when it was just dungeons but now that it's affixes too it's too small, and it can't be expanded

Image

src/gui/profile_editor/sigils_tab.py
```diff
@@ -75,34 +85,36 @@ def setup_ui(self):
...
-        remove_condition_btn = QPushButton("Remove Condition")
-        remove_condition_btn.clicked.connect(self.remove_selected)
-        condition_btn_layout.addWidget(remove_condition_btn)
+        form_layout.addRow("Affix:" if self.kind == "affix" else "Dungeon:", self.sigil_name_combo)
+
         layout.addLayout(form_layout)
-        layout.addLayout(condition_btn_layout)
-        layout.addLayout(title_layout)
-        layout.addWidget(self.condition_list)
+        if self.kind != "affix":
```
cjshrader: Unless I don't understand this code, the SigilConditionModels apply to both affixes and dungeons.

src/gui/profile_editor/sigils_tab.py
```diff
@@ -196,6 +208,19 @@ def create_form(self):
         self.priority_combobox.setMaximumWidth(150)
         self.priority_combobox.currentIndexChanged.connect(self.update_priority)
         self.general_form.addRow("Priority:", self.priority_combobox)
+
+        self.rarity_line_edit = QLineEdit()
```
cjshrader, line 211: Do you think rarity should work like this instead of being consistent with the rest? ie should we allow people to "whitelist" mythics and "blacklist" commons. Personally I've always found the sigil system to be too complex but at the same time this is different than how we have it

src/item/filter.py
```diff
@@ -196,13 +200,25 @@ def _check_cosmetic(item: Item) -> FilterResult:
         res.matched.append(MatchedFilter("Cosmetics"))
         return res
 
+    def _get_sigil_rarity(self, sigil_affixes) -> ItemRarity | None:
+        rarity_map = Dataloader().affix_sigil_dict_all["rarities"]
+        for affix in sigil_affixes:
+            if affix.name in rarity_map:
+                return ItemRarity(rarity_map[affix.name].lower())
+        LOGGER.debug(f"Could not resolve sigil rarity from affixes: {[a.name for a in sigil_affixes]}")
```
cjshrader: I'd think instead of figuring out the rarity here, we should figure it out when we generate the item itself. The rarity field of the item should be populated.

I'm also ok with this being a warning log. I've reworked a lot of how we pull data from d4data, and then chrizzo has continued that, and now our data from d4data is generally very accurate. I'd like to know if we're missing a rarity here.

tests/gui/importer/test_mobalytics.py
```diff
@@ -29,8 +29,6 @@
     "https://mobalytics.gg/diablo-4/profile/screamheart/builds/15x-thrash-out-of-date",
     # This one has an item type for the weapon
     "https://mobalytics.gg/diablo-4/builds/druid-zaior-pulverize-druid",
-    # This has two rogue offhand weapons
-    "https://mobalytics.gg/diablo-4/builds/rogue-efficientrogue-dance-of-knives?ws-ngf5-1=activeVariantId%2Ca2977139-f3e2-4b13-aa64-82ba69972528",
```
cjshrader, line 29: Why was this one removed? Looks like the link still works but the data could be outdated

README.md
```diff
@@ -335,6 +339,47 @@ Affixes:
 Affix names are lower case and spaces are replaced by underscore. You can find the full list of names
 in [assets/lang/enUS/affixes.json](assets/lang/enUS/affixes.json).
 
+### Filtering on rarity
+
+You can restrict an affix rule to one or more item rarities using the `rarity` key. This is useful for crafting: you can
+keep a common/magic/rare base that matches a rule while ignoring the equivalent legendary, which you would not craft on.
+
+- A rule with no `rarity` set matches all rarities, so existing rules behave exactly as before.
+- `rarity` accepts a single value (`rarity: rare`) or a list (`rarity: [common, magic, rare]`).
```
cjshrader: I think this block is too verbose. Some of this is just explaining how basic filtering works, and explaining things like rarity is case-insensitive isn't really relevant as we're very close to a fully-profile-editor world.

README.md
```diff
@@ -577,6 +622,24 @@ Sigils:
 
 </details>
 
+You can also gate all sigils by rarity using the top-level `rarity` key. Diablo 4 does not expose a rarity on sigils, so
+it is derived from the sigil's affixes via the rarities map in [assets/lang/enUS/sigils.json](assets/lang/enUS/sigils.json).
+
+- An empty/absent `rarity` keeps all sigil rarities (the gate is opt-in).
```
cjshrader: Once again just too verbose. I think a single sentence explaining it's applied before blacklist./whitelist is fine, if we keep that. The fact that we debug log an "unresolved lookup" (which is too technical for the readme) is not relevant to the user.

cjshrader (changes requested):

tests/gui/importer/test_mobalytics.py
chrisoro: The data was trash and I cba fixing it

src/gui/profile_editor/sigils_tab.py
chrisoro: Yup, couldn't test it on mac. I made it bigger + removed fixed size. I can now be resized.
default:
{B5EACA32-DDE3-4FFD-9C07-437CD9D5DF4D}
resized if you want:
{5E4C29D2-A926-446F-A7BB-8C43C0AF6256}

src/gui/profile_editor/sigils_tab.py
chrisoro: yea, editor now shows for both dungeon and affix entries in sigils_tab.py. so condition applies to both, ui now matches model/filter behavior.

src/item/filter.py
chrisoro: sure, didnt know that existed. sigil rarity is now populated during item parse in read_descr_tts.py from sigils.json rarity map based on data. I left it as a fallback though

README.md
chrisoro: better? if not just edit it how you want it :)

README.md
chrisoro: better? if not just edit it how you want it :)

cjshrader (approved):

src/gui/models/dialog.py
```diff
@@ -287,23 +299,29 @@ def __init__(self, whitelist_sigils: list[str], blacklist_sigils: list[str], par
         self.blacklist_sigils = blacklist_sigils
 
         self.setWindowTitle("Create Sigil")
-        self.setFixedSize(300, 150)
+        self.setMinimumSize(420, 220)
```
cjshrader, line 303: Everything is good except one little thing. We should save the size the user selects in QSettings just like we do with all the other windows

cjshrader (changes requested):

src/gui/models/dialog.py
chrisoro: aight, will change it in the aftermoon

src/item/filter.py
cjshrader, line 256: I missed this on the first pass but if res.keep is false and rarity is Mythic we want to keep it anyway. There's an example coming of this in seals/charms, but a similar example is here:

https://github.com/d4lfteam/d4lf/pull/786/changes#diff-5dad47261dae66ac6bf9f463ddc0085e3e4d36d22cf140ea15528e7aeee1b7a9L247

We want it to be after the normal filtering because if it matches a profile we want to show that. How I did it in Tributes is actually wrong. 

Here's the example for seals/charms: https://github.com/d4lfteam/d4lf/pull/763/changes#diff-5dad47261dae66ac6bf9f463ddc0085e3e4d36d22cf140ea15528e7aeee1b7a9R319

cjshrader (changes requested):

src/item/filter.py
cjshrader: Those links don't seem to work but it's line 319 of filter.py for the second example

src/gui/models/dialog.py
chrisoro: fixed

src/item/filter.py
chrisoro: fixed

cjshrader: You appear to have changed it for tributes, which is fine, but it's sigils that needed this added. Tributes were fine.

cjshrader (changes requested):

cjshrader (approved):

PR #787, opened 2026-06-16
(BugFix) Handle item mid-transition gracefully in vision mode. Report issues if this causes vision mode to not load too often

### LINKED ISSUE #697: Error in vision mode (Could not detect item_seperator_short)
I got this today when I entered a dungeon and immediately teleported to Den.
The problem might be that scene is already transitioning and my mouse hovered an item.

[log_2026_05_08_12_22_41.txt](https://github.com/user-attachments/files/27520115/log_2026_05_08_12_22_41.txt)

Image

chrisoro:
fixes https://github.com/d4lfteam/d4lf/issues/697

### DIFF
--- a/src/scripts/vision_mode_with_highlighting.py
+++ b/src/scripts/vision_mode_with_highlighting.py
@@ -372,6 +372,11 @@ class VisionModeWithHighlighting:
                                 item_descr_with_loc = item_descr
                             else:
                                 item_descr_with_loc = src.item.descr.read_descr_tts.read_descr_mixed(cropped_descr)
+                            if item_descr_with_loc is None:
+                                # Item was likely mid-transition and could not be read; skip without drawing
+                                LOGGER.debug("Item could not be read (likely in transition); skipping draw")
+                                self.request_clear()
+                                return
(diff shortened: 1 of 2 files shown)


src/scripts/vision_mode_with_highlighting.py
```diff
@@ -372,6 +372,11 @@ def evaluate_item_and_queue_draw(self, item_descr: Item):
                                 item_descr_with_loc = item_descr
                             else:
                                 item_descr_with_loc = src.item.descr.read_descr_tts.read_descr_mixed(cropped_descr)
+                            if item_descr_with_loc is None:
```
cjshrader: What is the end result here? I think ideally we should just loop again and give it a few tries to let the item come on screen fully.

Does this just end up clearing the overlay and that's it and nothing happens? That's how I read it. That'd be confusing.

chrisoro: Yup, clear overlay and print log message. I'll change it to just skip. We still do need this if or else we plug None into further processing and it'll result in the mentioned (bad) log message

cjshrader (approved):

PR #788, opened 2026-06-17
fix: apply log level changes without restart

### LINKED ISSUE #768: Changing log level to debug in settings does not automatically change the log level without a restart
Previously, if you changed log level, all future logs would be of the new log level without restart. Now you need to restart. This is a result of the new settings overhaul, something got missed.

chrisoro:
fixes https://github.com/d4lfteam/d4lf/issues/768


cjshrader (approved):
This appears to work, thank you. It's a small thing that no one will notice but it's a bit cleaner now

PR #789, opened 2026-06-17
Fix uc folder conflicts in tests

chrisoro:


cjshrader:
Oops, I actually removed this uc stuff in a previous PR for the same reason. I don't think we *need* it right now. If we add it back we'll need these changes again.

PR #794, opened 2026-06-26
(Dev) Centralized profile handling code into one location

chrisoro:


cjshrader (approved):
My quick tests looked good

PR #795, opened 2026-06-26
(Dev) Cleaned up TTS processing flow

chrisoro:


cjshrader (approved):
I was aware there was some duplicated code here and this does seem to have cleaned up some of it so thank you

PR #796, opened 2026-06-29
(Profile Editor) Updates to handle charms and seals

chrizzocb:
Summary of Changes
1. Data Models & Validation

profile_models.py
:
Moved the set field and its validation to the shared base class _BaseSealOrCharmFilterModel. This enables both CharmFilterModel and SealFilterModel to support set-name binding as well as the mutual exclusion constraint against unique_aspects.
2. User Interface & Tab Integration

profile_editor.py
:
Integrated the CharmsTab and SealsTab into the main Profile Editor window with lazy loading wired up inside tab_changed.

charms_tab.py
:
[NEW] Created the Charm editor tab, supporting Rarities, Sets, Min Greater Affixes (with auto-sync), Unique Aspects, and Affix Pools.

seals_tab.py
:
[NEW] Created the Seal editor tab, mirroring the Charm editor architecture.
3. Affix Selection & Set-Bound Affixes

affixes_tab.py
:
Added a Set combobox selector in front of the Affix name combobox inside AffixWidget when editing charms or seals.
Selecting a Set (e.g. sescherons_fury) filters the Affix combobox to show only that Set's suffixes without the prefix (e.g., "damage" instead of "sescherons fury damage"). Selecting (None) displays normal general affixes. The fully qualified key name is still written back to the model correctly.
Fixed a parent widget resolution bug: Now passing the parent group editor (self) during AffixPoolWidget instantiation inside add_affix_pool_item. This ensures _affix_dict_for_widget correctly walks up the parent tree to locate the SealFilterModel / CharmFilterModel context and load the appropriate dictionary (seal_affix_dict / charm_affix_dict) instead of falling back to item affixes.
4. Dialogs

dialog.py
:
[NEW] Created CreateCharmOrSeal dialog to handle named filter creation.
[NEW] Created SetPicker dialog to allow multi-selection of set names.
5. Test Robustness

loader.py
:
Fixed a test suite mock error: _remove_defunct_model_keys now dynamically verifies the presence of model_fields via getattr to prevent crashes when mocked model classes (MagicMock) are active in parallel test execution.
Verification
uv run prek run -a — All formatting and Ruff linting pre-commit checks pass successfully. ✅
uv run pytest tests/item/filter/filter_test.py — All 76 item filter tests (including charm/seal matching) pass successfully. ✅


src/config/loader.py
```diff
@@ -89,7 +89,10 @@ def _remove_defunct_model_keys(self) -> bool:
             if section not in self._parser:
                 continue
 
-            valid_keys = type(model).model_fields
+            valid_keys = getattr(type(model), "model_fields", None)
```
cjshrader: Any time I see something replaced with getattr(...) I assume it is wrong. Please revert or explain why this was needed. This change is not related to seals or charms

src/config/profile_models.py
```diff
@@ -268,6 +268,7 @@ class _BaseSealOrCharmFilterModel(BaseModel):
     affix_pool: list[AffixFilterCountModel] = Field(default=[], alias="affixPool")
     min_greater_affix_count: int = Field(default=0, alias="minGreaterAffixCount")
     rarities: list[ItemRarity] = Field(default=[], validation_alias="rarity", serialization_alias="rarity")
+    set: list[str] = Field(default=[], alias="set")
```
cjshrader: Set is not a valid field on this model, which is why it wasn't already there. It's on the charm. If you're doing profile editor updates and have to change the model it's a red flag

src/config/profile_models.py
```diff
@@ -288,17 +294,8 @@ def unique_aspects_must_be_unique(self) -> _BaseSealOrCharmFilterModel:
 
         return self
 
-
-class CharmFilterModel(_BaseSealOrCharmFilterModel):
```
cjshrader, line 294: Yeah this makes no sense. Let me know if I'm wrong.

cjshrader (changes requested):
I only reviewed loader/profile_models but no reason to review the rest until those changes are explained or addressed

src/config/profile_models.py
chrizzocb: take a look at the update comment

src/config/profile_models.py
chrizzocb: take a look at the update comment

src/config/loader.py
chrizzocb: take a look at the update comment

cjshrader (changes requested):
Image

A charm affix can't have a set, this split doesn't make any sense.

It works well for seals though, I like it. Can we change (None) to (No Set Selected) to make it more clear what that drop down is?

Image

Can we change the unique aspect for seals to only display unique aspects that start with "seal_of"?

Image

src/gui/profile_editor/charms_tab.py
```diff
@@ -0,0 +1,544 @@
...
+                group = CharmGroupEditor(item)
+                self.item_names.append(item_name)
+                self.tab_widget.addTab(group, item_name)
+                self.charms_model.append(item)
+            return
+
+    def close_tab(self, index):
+        self.item_names.pop(index)
+        self.tab_widget.removeTab(index)
+        self.charms_model.pop(index)
```
chrisoro: Bug: pops `self.charms_model[index]` by tab position, but `DynamicCharmFilterModel = RootModel[dict[str, CharmFilterModel]]` can hold multiple `item_name` keys per list entry. One tab is created per `item_name` (see loop below in `setup_ui`), so tab count can exceed `charms_model` length. Closing any tab after a multi-key group entry pops the wrong model, silently corrupting unrelated charm data. Same bug in `remove_item_type` (line 498) and mirrored in `seals_tab.py:449/459`.

src/gui/profile_editor/charms_tab.py
```diff
@@ -0,0 +1,544 @@
...
+CHARMS_TABNAME = "Charms"
+
+
+def _set_summary(sets: list[str]) -> str:
+    if not sets:
+        return "No sets selected"
+    return ", ".join(sets)
+
+
+class CharmGroupEditor(QWidget):
```
chrisoro: `CharmGroupEditor` is ~95% identical to `SealGroupEditor` in seals_tab.py (rarities, min-greater-affix + auto-sync, affix-pool management, greater-count-label helpers), and both re-implement logic that already exists nearly verbatim in `affixes_tab.py`'s `AffixGroupEditor`. Suggest extracting a shared base class/mixin instead of copy-pasting two ~500-line files. Already diverging: `seals_tab.py`'s `_create_readonly_line_edit` has a stray blank line the charm copy doesn't — proof a fix to one won't propagate to the other.

src/gui/profile_editor/affixes_tab.py
```diff
@@ -879,12 +956,20 @@ def create_value_input(self):
...
         affix_dict = self.get_affix_dict()
-        reverse_dict = {v: k for k, v in affix_dict.items()}
-        affix_name = reverse_dict.get(current_text or self.name_combo.currentText())
-        if affix_name is None:
-            return
-        self.affix.name = affix_name
+        text = current_text or self.name_combo.currentText()
+
+        if is_seal_charm:
+            reverse_dict = {v: k for k, v in self.filtered_affixes.items()}
```
chrisoro, line 972: When the selected Set filters to zero matching affixes (`on_set_changed` → `name_combo.count() == 0`, line 882), `update_name` is never called, and here the lookup silently no-ops if `text` isn't in `filtered_affixes` — `self.affix.name` keeps pointing at the previous set's affix key while the combo shows nothing/a different value. That stale mismatch gets written back to the saved profile.

src/gui/profile_editor/charms_tab.py
```diff
@@ -0,0 +1,544 @@
...
+        self.min_greater.setToolTip(
+            "Minimum number of checked affixes that must be Greater Affixes.\n"
+            "0 = Accept items even without GAs (for leveling)\n"
+            "1-4 = At least this many checked affixes must be GA"
+        )
+        self.min_greater.valueChanged.connect(self.update_min_greater_affix)
+
+        self.auto_sync_checkbox = _create_auto_sync_checkbox()
+        self.auto_sync_checkbox.setChecked(
+            self.settings.value(f"auto_sync_ga_{self.item_name}", defaultValue=False, type=bool)
```
chrisoro: QSettings key `f"auto_sync_ga_{self.item_name}"` isn't namespaced by charm vs. seal. A charm and seal sharing the same name (e.g. both `"MyBuild"`) share persisted auto-sync state across tab types. Same in `seals_tab.py:102/310`.

src/gui/models/dialog.py
```diff
@@ -681,3 +685,106 @@ def clear_selection(self):
...
+            return
+        if self.name_input.text() in self.item_list:
+            QMessageBox.warning(self, "Warning", "Name already exists")
+            return
+        super().accept()
+
+    def get_value(self):
+        item_name = self.name_input.text()
+        affix_dict = Dataloader().charm_affix_dict if self.is_charm else Dataloader().seal_affix_dict
+        default_affix = AffixFilterModel(name=next(iter(affix_dict.keys())), value=None)
```
chrisoro: `next(iter(affix_dict.keys()))` raises uncaught `StopIteration` if `charm_affix_dict`/`seal_affix_dict` is empty (e.g. Dataloader fails to load charm/seal affix data). Crashes "Create Charm/Seal" instead of showing an error. Same pattern duplicated in `charms_tab.py:304` and `seals_tab.py:265` (`add_affix_pool`).

src/gui/profile_editor/charms_tab.py
```diff
@@ -0,0 +1,544 @@
...
+        self.setLayout(main_layout)
+
+        QTimer.singleShot(100, self.affix_pool_container.expand)
+
+    # --- Sets (charm-specific) ---
+
+    def refresh_set_summary(self):
+        self.set_line_edit.setText(_set_summary(self.config.set))
+
+    def edit_sets(self):
```
chrisoro: Set-vs-unique_aspect mutual exclusion is enforced only here in the UI (and duplicated in `AffixGroupEditor`), not as a model-level pydantic validator on `CharmFilterModel`/`SealFilterModel`. A profile edited outside this dialog path (script, hand-edited YAML) can end up with both fields populated with no validation error.

src/gui/profile_editor/charms_tab.py
```diff
@@ -0,0 +1,544 @@
...
+    DeleteItem,
+    MinGreaterDialog,
+    MinPercentDialog,
+    RarityPicker,
+    SetPicker,
+    rarity_summary,
+)
+from src.gui.profile_editor.affixes_tab import UNIQUE_ASPECTS_TITLE, AffixPoolWidget, AffixWidget, UniqueAspectWidget
+
+LOGGER = logging.getLogger(__name__)
```
chrisoro: `LOGGER` is never used in this file (or seals_tab.py) — dead setup. Also `_create_auto_sync_checkbox` (line 531) does a local `from PyQt6.QtWidgets import QCheckBox` with a `noqa: PLC0415` suppression instead of adding `QCheckBox` to the existing top-of-file import block that already imports a dozen other widgets.

chrisoro (commented):
Automated review: found one correctness bug (tab/model index mismatch), a stale-state bug in the set-scoped affix combo, a QSettings key collision, an uncaught StopIteration risk, and significant duplication (charms_tab.py/seals_tab.py mirror each other and re-implement AffixGroupEditor logic).

src/gui/profile_editor/charms_tab.py
chrizzocb: take a look at the next update.

src/gui/profile_editor/charms_tab.py
chrizzocb: take a look at the next update.

src/gui/models/dialog.py
chrizzocb: take a look at the next update.

src/gui/profile_editor/charms_tab.py
chrizzocb: take a look at the next update.

src/gui/profile_editor/affixes_tab.py
chrizzocb: take a look at the next update.

src/gui/profile_editor/charms_tab.py
chrizzocb: take a look at the next update.

src/gui/profile_editor/charms_tab.py
chrizzocb: take a look at the next update.

src/config/profile_models.py
```diff
@@ -112,8 +112,12 @@ def value_and_percent_are_mutually_exclusive(self) -> AffixFilterModel:
 class AffixFilterCountModel(BaseModel):
     model_config = ConfigDict(extra="forbid", populate_by_name=True)
     count: list[AffixFilterModel] = []
-    max_count: int = Field(default=sys.maxsize, alias="maxCount")
-    min_count: int = Field(default=0, alias="minCount")
+    max_count: int = Field(
```
cjshrader: I doubt these changes were needed. What issue was being addressed? I think it can be solved on the other end and not here.

cjshrader (changes requested):

cjshrader (changes requested):
When I try to open the rarity or set picker for seals or charms I get this error:

Traceback (most recent call last):
  File "C:\Users\cjshr\code\d4lf\src\gui\profile_editor\charms_seals_group_tab.py", line 297, in edit_rarities
    dialog = RarityPicker(self.config.rarities, parent=self)
TypeError: RarityPicker.__init__() got multiple values for argument 'parent'

cjshrader (dismissed):
Everything is looking good to me, I think we're ready to release this one to the wild

PR #800, opened 2026-07-02
(Importer) Add InfinityBuilds support for importing builds. Paragon support to come later if at all

chrisoro:


cjshrader (changes requested):
I'd like to see paragon support added now if possible, unless you intend to do it next. Mostly if we don't add it now or very soon then no one ever will.

Also this instructions screen was already too big and now it's half the screen, can we find a way to minimize it? Just slapping a scrollbox on it is fine and giving it a much smaller max height

Image

chrisoro:
Won't add paragon in here, this will take forever since the data isn't available via the api and binary encoded. reverse engineering will take a while, so lets get this out asap. paragon is a fringe use case anyways and the omittion is logged and noted in readme

cjshrader (approved):

PR #805, opened 2026-07-03
(Importer) Add InfinityBuilds paragon

chrisoro:


cjshrader (approved):
I did a quick check and it all looked good, both in import and in-game

PR #806, opened 2026-07-03
(BugFix) A variety of bug fixes and improvements

chrisoro:
1. Mythic flash-then-vanish - vision_mode_with_highlighting.py: added bounded retry (3 attempts, re-grab + re-crop) around the single-shot read_descr_mixed() call, which previously had zero retry margin for a transient bullet-template miss despite the code's own error message acknowledging such misses are often temporary.
  2. Masterworked/tempered unique losing affix markers - likely same root cause as 1. (transient bullet-detection miss nukes the whole draw with no retry), covered by the same fix. Also fixed a real, independent bug found along the way: AffixType.tempered was defined but never assigned anywhere in the codebase - tempered bullets fell through to AffixType.normal, silently making filter.py's "exclude tempered from matching" logic a no-op.
  3. Protective ancestral filter not working - the setting checked "has a greater affix" as a stand-in for "is ancestral," but those aren't the same thing (an ancestral item can roll zero GAs). The tooltip literally contains the word "Ancestral" and the parser was reading then discarding it. Added a real Item.is_ancestral field and wired it through.
  4. Armory charm affix marker shift - charms/seals used a hardcoded fixed line-index for where affixes start, unlike every other item type which anchors dynamically on tooltip text. Any inserted line (e.g. an Armory banner) shifted everything by one. Now anchors on the "Item Power" line like the rest of the parser already does elsewhere, so it's robust to extra inserted lines regardless of cause.


cjshrader (approved):
I'm not a big fan of trying to fix a bug that we haven't reproduced ourselves. schoof did not provide nearly enough information to address the flickering situation, I actually suspect it might be a resolution issue over everything else as most people don't play at 1920x1080.

That being said I think the retry solution is just generically useful so I'm happy to have it and who knows maybe it does fix the problem.

PR #807, opened 2026-07-03
(Dev) Consolidate all Tkinter overlays onto one shared UI thread

chrisoro:


cjshrader (approved):
Once again didn't see any problems in my testing, seems ok. I had every overlay up at once without issue.

PR #811, opened 2026-07-04
(Dev) Extract importer ImportPipeline skeleton

chrisoro:
Pull the shared guide-importer assemble/name/save/add-to-profiles flow into ImportPipeline, keep the four site entry points as thin wrappers over extracted build data, unify Maxroll affix deduplication with the shared pipeline behavior, and delete the dead diablo_trade importer and test.


cjshrader (changes requested):
I actually don't know if this PR caused it but somewhere along the way we lost logging in the importers in the GUI. For this importer I'd expect to see these logs, which I do see in the console:

```
Loading https://maxroll.gg/d4/planner/wfd4010h#3
Created profile C:\Users\cjshr\.d4lf\profiles\ringtest.yaml
Finished
```

Instead I only see what's in the screenshot, which is of course confusing.

Image

cjshrader (approved):

PR #813, opened 2026-07-04
The Profile tab now uses a new non-Qt ProfileSession seam

chrisoro:


cjshrader (approved):
I couldn't find any errors, seems good

PR #814, opened 2026-07-04
(Dev) Refactor list-based profile editor tabs into RuleListTab

chrisoro:


cjshrader (approved):
I discovered a small issue with tributes in the profile editor so I went ahead and fixed that, but I don't believe it was caused by these changes. I think it actually was caused by a pydantic update.

PR #820, opened 2026-07-05
(ProfileEditor) Tribute tab now has a rarity picker just like the other tabs that have rarity

### LINKED ISSUE #819: Tributes tab of Profile Editor is inconsistent with how other tabs handle rarity
Affixes and Sigils have one field for rarities and rarities are saved as one list in the profile (examples below):

Image

Image

Tributes work differently by having you select each individual rarity. They should just be consistent with the other two tabs.

Image

Additionally this saves like this:

```yaml
tributes:
- {name: tribute_of_heritage}
- {name: mythic_tribute_of_armaments}
- rarities: [common]
- rarities: [magic]
```

I don't even know if this will work, but it should be just one single list even if it does work.

chrisoro:
fixes https://github.com/d4lfteam/d4lf/issues/819


README.md
```diff
@@ -705,31 +711,34 @@ names in [assets/lang/enUS/sigils.json](assets/lang/enUS/sigils.json).
...
+constraints are ANDed together (same semantics as affixes/sigils): when both lists are non-empty, a tribute must match
+both. An empty list means that dimension is unconstrained. If no Tribute filter is provided, all Tributes are kept.
 
 Mythic tributes are always kept no matter what.
 
 <details><summary>Config Examples</summary>
 
 ```yaml
-# Keeps tribute_of_mystique and all legendary and unique tributes
+# Keeps only tribute_of_mystique when it is legendary or unique
```
cjshrader: This isn't correct functionality, is this accurate to how it works? The original comment was how it should work. You should be able to say "I want to keep tribute of mystique and also keep all legendary or unique tributes"

README.md
```diff
@@ -705,31 +711,34 @@ names in [assets/lang/enUS/sigils.json](assets/lang/enUS/sigils.json).
 
 ### Tributes
 
-Tributes are defined by the top-level key `Tributes`. It contains a list of either tribute names or rarities you want
-to keep. Any not in the list are not kept. If no Tribute filter is provided, all Tributes will be kept.
+Tributes are defined by the top-level key `Tributes` as a single object with `name` and `rarity` lists. These
```
cjshrader: I don't think users really follow words like "Semantics" and ANDed together

README.md
```diff
@@ -86,9 +86,15 @@ The TTS dll (`saapi64.dll`) must be signed for Diablo 4 to pick it up. The `inst
 - Download the signtool needed to add a local signature to the dll
 - Runs the signtool and signs the dll
 
-If you prefer running it from a terminal, you can run `.\install_dll.cmd`.
+If you prefer running it from a terminal:
 
-For very advanced users that don't want to automatically download signtool.exe, you can run `.\install_dll.cmd -signtool_path "<full path to signtool.exe>"`
+- release zip: `.\install_dll.cmd`
+- source checkout: `.\tts\install_dll.cmd`
```
cjshrader: I think if the user is doing a source checkout they can figure this out themselves. The readme is already long enough. Honestly I'm fine to remove the signtool stuff too.

README.md
cjshrader: Looking at the code I *think* it works how the original comment said still and it's just this comment that's wrong

cjshrader (changes requested):

README.md
chrisoro: as discussed via discord, comment is matching code. I changed logic to be the same as for everything else

cjshrader (changes requested):
My main ask was that Tributes would use the same rarity picker as the other tabs in the profile editor, where you just check boxes of the rarities you want to keep instead of this system where you have to individually add every single rarity. I'd like it if we can also make that change

cjshrader (changes requested):
This is what my profile looks like:

tributes:
- rarities: [rare]
- rarities: [legendary]
- rarities: [mythic]

But it doesn't render correctly in the profile editor:

Image

If I change it to the much more sane:

tributes:
- rarities: [rare, legendary, mythic]

Then it does render correctly:

Image

The way it was before is unfortunately how the profile editor saved it before, which is part of the reason I asked for these changes

src/gui/profile_editor/tributes_tab.py
```diff
@@ -1,55 +1,112 @@
...
+        main_layout.setContentsMargins(0, 20, 0, 20)
+        main_layout.setAlignment(Qt.AlignmentFlag.AlignTop)
 
-class TributesTab(RuleListTab[TributeFilterModel]):
-    def __init__(self, tributes: list[TributeFilterModel] | None, parent=None):
-        super().__init__(tributes, parent)
-        self.tributes = self.items
-        self.tribute_list_widget = self.list_widget
+        label = QLabel(
+            "Add tribute names and tribute rarities you want to keep. Name and rarity constraints are ANDed together."
```
cjshrader: Let's not use terms like ANDed together, though I don't think this comment is even true

src/gui/models/activity_log_widget.py
```diff
@@ -435,8 +435,10 @@ def _get_profile_summary(self, path: Path) -> str:
                 summary.append(f"💎 Global Uniques: {len(model.global_uniques)}")
             if model.sigils:
                 summary.append("📜 Sigils: Included")
-            if model.tributes:
-                summary.append("🏆 Tributes: Included")
+            if model.tributes is not None:
```
cjshrader: Isn't model.tributes still enough? Isn't it set to None or an empty list if it doesn't exist?

cjshrader (changes requested):
Functionality wise it looked good in-game, I had no issues there. Thank you for doing this in such a way that it wasn't a breaking change.

src/gui/profile_editor/tributes_tab.py
chrisoro: leftover from v1

chrisoro:
{A01BAD6B-68F8-42CD-99CA-F4F4EB07B7C3}

src/gui/models/activity_log_widget.py
```diff
@@ -436,7 +436,9 @@ def _get_profile_summary(self, path: Path) -> str:
             if model.sigils:
                 summary.append("📜 Sigils: Included")
             if model.tributes:
-                summary.append("🏆 Tributes: Included")
+                tribute_rules = model.tributes if isinstance(model.tributes, list) else [model.tributes]
```
cjshrader: I wasn't clear enough earlier. Did this code need to change at all? Is it not enough to just say "model.tributes"? Or do we absolutely have to check if there are any rule.names and rule.rarities?

cjshrader (changes requested):
The pre-change tributes:

```
tributes:
- rarities: [rare]
- rarities: [legendary]
- rarities: [unique]
- {name: lesser_tribute_of_harmony}
```

Are still not being shown correctly in the GUI, only rare is shown and not the other items.

Image

src/gui/models/activity_log_widget.py
chrisoro: Gotcha, nope it was part of my initial implementation which we changed. so I'll remove it

chrisoro:
I'll just change it so that we take one list of names and rarities and use both data as "or". I'll transform it into a reusable object internally which is the best of both worlds I guess. You want the user interface to stay the same and I want it to be compliant to our architecture ;)

cjshrader (approved):
That solution makes a lot of sense, and when it's saved it is saved in the much more sane new format:

tributes:
  name: [lesser_tribute_of_harmony]
  rarities: [rare, legendary, unique]

PR #824, opened 2026-07-06
(Importer) Mythics are now imported as separate items with affixes again

### LINKED ISSUE #818: Additional reworks are needed for Mythics, specifically for importers
Reworks need to be done because Mythics are not what they used to be. 

First, the importers previously lumped all mythics together into one "Mythic" section with no affixes included. Since mythics are now just uniques but purple, I think we should import them as their own individual items with their own individual affixes so someone can check if the mythic they have has the stats they want.

There may be additional issues with mythics, I've barely had one in the game. A unit test needs to be added, here is the TTS string:

['BLOOD-MAD IDOL', 'Ancestral Mythic Unique Amulet', '900 Item Power', '230 All Resist (-7.4% Toughness)', '+180 Willpower +[150 - 180]', '+1,225 Armor [981 - 1,225]', '+10.0% Attack Speed [8.0 - 10.0]%', 'x24% Shadow Damage Multiplier [14 - 24]%', 'You are always Berserking but take 200%[x] increased damage as Burning over 8 seconds. While Burning, Berserking grants an additional 195%[x] [156 - 195]% increased damage.', 'The relic was fashioned for the broken-hearted and the battle-mad, for those whose rage has nowhere left to dwell. - Notes from Sister Octavia on the Burnt Knights', 'Requires Level 70. Account Bound. Unique Equipped', 'Sell Value: 1 Gold', 'Tempers: 3/3', 'Right mouse button']

And here is the parsed item:

Item(affixes=[Affix(loc=None, max_value=180.0, min_value=150.0, name='willpower', text='+180 Willpower +[150 - 180]', type=<AffixType.normal: 3>, value=180.0), Affix(loc=None, max_value=1225.0, min_value=981.0, name='armor', text='+1,225 Armor [981 - 1,225]', type=<AffixType.normal: 3>, value=1225.0), Affix(loc=None, max_value=10.0, min_value=8.0, name='attack_speed', text='+10.0% Attack Speed [8.0 - 10.0]%', type=<AffixType.normal: 3>, value=10.0), Affix(loc=None, max_value=24.0, min_value=14.0, name='shadow_damage_multiplier', text='x24% Shadow Damage Multiplier [14 - 24]%', type=<AffixType.normal: 3>, value=24.0)], aspect=Aspect(name='blood-mad_idol', loc=None, min_value=None, max_value=None, text='You are always Berserking but take 200%[x] increased damage as Burning over 8 seconds. While Burning, Berserking grants an additional 195%[x] [156 - 195]% increased damage.', value=200.0), codex_upgrade=False, cosmetic_upgrade=False, inherent=[], is_ancestral=True, is_in_shop=False, item_type=<ItemType.Amulet: 'amulet'>, name='blood-mad_idol', original_name='BLOOD-MAD IDOL', power=900, rarity=<ItemRarity.Mythic: 'mythic'>, seasonal_attribute=None, set=None)

It actually looks like we got lucky and we 100% correctly parse mythics still so there may be no additional issues. There could be a discussion around removing the "Always keep mythics" rule but I don't think so, if someone finds a mythic that's still pretty rare and if they upgrade an item to one that was a very intentional decision so it makes sense to me to still always keep every mythic we find.

chrisoro:
fixes https://github.com/d4lfteam/d4lf/issues/818

### DIFF
--- a/tests/item/read_descr_tts_equipment_items_test.py
+++ b/tests/item/read_descr_tts_equipment_items_test.py
@@ -1,0 +1,31 @@
+import platform
+from typing import TYPE_CHECKING
+
+import pytest
+
+from src.item.data.item_type import ItemType
+from tests.item.read_descr_tts_cases_bloodied_items import items as bloodied_items
+from tests.item.read_descr_tts_cases_horadric_and_equipment import items as horadric_and_equipment_items
+from tests.item.read_descr_tts_cases_mythic_items import items as mythic_items
+from tests.item.read_descr_tts_cases_sanctified_and_sigils import items as sanctified_and_sigil_items
+from tests.item.read_descr_tts_cases_unique_gloves import items as unique_items
+
+if TYPE_CHECKING:
+    from src.item.models import Item
+
+if platform.system() == "Windows":
+    import src.tts
+    from src.item.descr.read_descr_tts import read_descr
+
+ALL_ITEMS = [*mythic_items, *unique_items, *sanctified_and_sigil_items, *bloodied_items, *horadric_and_equipment_items]
+
+NON_EQUIPMENT_TYPES = {ItemType.Sigil, ItemType.EscalationSigil, ItemType.HoradricSeal, ItemType.Charm}
+items = [item for item in ALL_ITEMS if item[1].item_type not in NON_EQUIPMENT_TYPES]
+pytestmark = pytest.mark.skipif(platform.system() != "Windows", reason="Requires Windows TTS modules")
+
+
+@pytest.mark.parametrize(("input_item", "expected_item"), items)
+def test_equipment_items(input_item: list[str], expected_item: Item):
+    src.tts.LAST_ITEM = input_item
+    item = read_descr()
+    assert item == expected_item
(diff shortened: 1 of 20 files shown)


tests/item/read_descr_tts_equipment_items_test.py
```diff
@@ -0,0 +1,31 @@
...
+
+ALL_ITEMS = [*mythic_items, *unique_items, *sanctified_and_sigil_items, *bloodied_items, *horadric_and_equipment_items]
+
+NON_EQUIPMENT_TYPES = {ItemType.Sigil, ItemType.EscalationSigil, ItemType.HoradricSeal, ItemType.Charm}
+items = [item for item in ALL_ITEMS if item[1].item_type not in NON_EQUIPMENT_TYPES]
+pytestmark = pytest.mark.skipif(platform.system() != "Windows", reason="Requires Windows TTS modules")
+
+
+@pytest.mark.parametrize(("input_item", "expected_item"), items)
+def test_equipment_items(input_item: list[str], expected_item: Item):
```
cjshrader, line 28: I don't mind this change. Probably need to clean up the renamed files further because they weren't really separated by specific functionalities like this but I'm fine to go forward with this as-is

cjshrader (approved):
I did a couple of tests and everything looked good. Something I missed is we were already handling maxroll this way, which implies to me they just report these items as normal uniques to us anyway. But it's good to go ahead and make the change in case they do start reporting them as mythics in the future.

PR #825, opened 2026-07-06
(Dev) Add GUI only mode for testing on non-Windows OS

chrisoro:
Also replaced mouse and keyboard input (both outdated and win only) with generic pynput


cjshrader (changes requested):
When you press F11 to filter items, the mouse no longer travels from point to point but instead teleports from item to item. This does not always allow Diablo enough time to load the new item and you'll end up marking the previous items as junk or favoriting it instead.

The mouse cursor needs to have some travel time. I actually did a lot of trial and error previously so whatever travel time I had before needs to be kept as close to accurate as possible.

chrisoro:
Strange, on my pc it does work without any problems. Maybe it is a performance thing? Either way, I made it more similar to what is was before but without the fat dependency

cjshrader (changes requested):
Although the mouse still "teleports" from item to item, it looks like your changes added enough of a delay that I wasn't able to reproduce the problem. However, if we could get closer to the old solution I'd prefer that because it's a little less clear if anyone is looking out for bots that this is effectively a bot. Moving the mouse from one item to another is what a human does, but teleporting the mouse is what a bot does.

However in testing I also discovered hotkeys that use shift or ctrl to activate, like shift + f11, no longer work.

chrisoro:
Try again, since it seems to differ between system. I'm only on my mac today, so I can only verify the key fix. Mouse should not be very similar to before

cjshrader (changes requested):
Mouse movement is good now, thank you.

I'm still having issues with the hotkeys. F11 works fine but shift+f11 does nothing except move the cursor to the middle of the screen.

What's strange is it logs the following after pressing shift+F11:

```
Stopping Vision Mode
Running loot filter
Stopping filter or move process
Starting Vision Mode
```

So it's doing *something* it's just not executing the actual filtering. I'm unclear why that might be.

I tried the command on the current main branch just to make sure the problem wasn't somewhere else but that worked fine.

chrisoro:
As explained, this one is hard for me as I still don't have access to my Windows pc for a few more days.

I think the bug was that both hotkeys fired and thus one script got killed immediately. If that was really the case, it should be guarded now.

cjshrader (approved):
Appears to be working well now for me

PR #826, opened 2026-07-06
(Dev) Replaced .pre-commit-config.yaml with prek.toml

chrisoro:


cjshrader:
Just to make sure I understand, this does not replace pre-commit but is just a different way of setting it up? So the readme instruction of running pre-commit install is still accurate and this will still prevent me from making a commit if there are prek issues?

chrisoro:
Yea no functionality changes, only the new toml format. I also changed readme a bit because it was very confusing

cjshrader (approved):

PR #828, opened 2026-07-07
(Vision Mode With Highlighting) Improve vision mode with tts info

chrisoro:
Improved accuracy from 91 to 100% on a local testset while improved speed from 70ms to 20ms on my pc.

<details open><summary><h3>Greptile Summary</h3></summary>

This PR updates vision mode with highlighting to combine TTS item parsing with tooltip geometry matching. The main changes are:

- TTS-only item parsing for vision mode evaluation.
- Separate geometry-based marker localization for matched affixes and aspects.
- Broader tooltip, separator, and bullet template coverage.
- Replay tools for full screenshots, cropped tooltips, and template matching diagnostics.
- Build artifact naming support for a version override.
</details>

<details open><summary><h3>Confidence Score: 5/5</h3></summary>

This PR appears safe to merge.

No new verified issues were identified in the changed runtime paths. The main behavior changes have focused tests for TTS parsing, tooltip detection, marker localization, and replay tooling.

No files require special attention.
</details>

<details><summary><h3><a href="https://www.greptile.com/trex">T-Rex</a> T-Rex Logs</h3></summary>

**What T-Rex did**
- - The vision\_changed tests were executed and logged; the main proof confirms the command, environment notes, full Pytest output, and that the test process exited with code 0.
- - A runtime replay was performed; the replay log shows the replay command, found=True, match\_count=1, and EXIT\_CODE: 0.
- - Visual proof files were generated to document the replay, including native replay input and output images as well as non-native exploratory visuals.
- - The uploaded artifacts provide evidence for the test run, the replay results, and the associated visuals.

<a href="https://app.greptile.com/trex/runs/14877925/artifacts"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/ViewAllArtifactsDark.svg?v=4"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/ViewAllArtifacts.svg?v=4">View all artifacts</picture></a>

<sub><a href="https://www.greptile.com/trex">T-Rex</a> Ran code and verified through T-Rex</sub>
</details>

<details><summary><h3>Important Files Changed</h3></summary>

| Filename | Overview |
|----------|----------|
| src/scripts/vision_mode_with_highlighting.py | Switches highlighting flow to TTS parsing plus separate cropped-tooltip marker localization and retry handling. |
| src/item/descr/read_descr_tts.py | Refactors TTS parsing to remove image-location coupling and adds variable affix counting for TTS-only vision mode. |
| src/item/descr/geometry_locator.py | Adds TTS-guided marker localization with diagnostics for affix and aspect bullets. |
| src/item/descr/texture.py | Generalizes separator and bullet template matching and adds tracing support for marker diagnostics. |
| src/item/find_descr.py | Reworks tooltip detection to return diagnostic matches, clip crop regions, and expose separator translation for cropped tooltips. |
| src/template_finder.py | Adds optional early-stop matching while preserving template order for multi-template stop-condition searches. |
| src/tools/replay_cropped_tooltip.py | Adds a replay/debug tool for cropped-tooltip marker matching with validation and annotated output. |
| src/tools/replay_full_screenshot.py | Adds full-screenshot replay tooling for tooltip detection diagnostics and crop export. |
| .github/actions/build_zip/action.yml | Adds an optional version override for ZIP artifact naming while retaining the package version fallback. |
| .github/workflows/build.yml | Passes the workflow commit SHA as the build ZIP/artifact version override. |

</details>

<details><summary><h3>Sequence Diagram</h3></summary>

<a href="#gh-light-mode-only">

```mermaid
%%{init: {'theme': 'neutral'}}%%
sequenceDiagram
participant TTS as TTS pipeline
participant Vision as VisionModeWithHighlighting
participant Filter as Item Filter
participant Detect as Tooltip Detection
participant Locator as Geometry Locator
participant Overlay as Tk Overlay

TTS->>Vision: LAST_ITEM update
Vision->>Vision: read_descr() parses Item from TTS
Vision->>Detect: find_descr_with_diagnostics(screenshot, item center)
Detect-->>Vision: cropped tooltip ROI + separator match
Vision->>Filter: should_keep(item)
Filter-->>Vision: keep/junk result + matched affixes/aspects
alt kept equipment
    Vision->>Locator: locate_affix_markers(cropped tooltip, Item, matches)
    Locator-->>Vision: reliable marker centers
    Vision->>Overlay: draw keep outline + matched markers
else no match or ignored
    Vision->>Overlay: draw no-match/empty outline
end
```

</a>
<a href="#gh-dark-mode-only">

```mermaid
%%{init: {'theme': 'base', 'themeVariables': {"darkMode": true, "background": "#0d1117", "primaryColor": "#21262d", "primaryTextColor": "#e6edf3", "primaryBorderColor": "#8b949e", "lineColor": "#8b949e", "textColor": "#e6edf3", "edgeLabelBackground": "#161b22", "actorBkg": "#21262d", "actorBorder": "#8b949e", "actorTextColor": "#e6edf3", "actorLineColor": "#8b949e", "signalColor": "#8b949e", "signalTextColor": "#e6edf3", "noteBkgColor": "#373320", "noteBorderColor": "#d4a72c", "noteTextColor": "#f0e6c0", "labelBoxBkgColor": "#21262d", "labelBoxBorderColor": "#8b949e", "labelTextColor": "#e6edf3", "loopTextColor": "#e6edf3", "activationBkgColor": "#30363d", "activationBorderColor": "#8b949e"}}}%%
sequenceDiagram
participant TTS as TTS pipeline
participant Vision as VisionModeWithHighlighting
participant Filter as Item Filter
participant Detect as Tooltip Detection
participant Locator as Geometry Locator
participant Overlay as Tk Overlay

TTS->>Vision: LAST_ITEM update
Vision->>Vision: read_descr() parses Item from TTS
Vision->>Detect: find_descr_with_diagnostics(screenshot, item center)
Detect-->>Vision: cropped tooltip ROI + separator match
Vision->>Filter: should_keep(item)
Filter-->>Vision: keep/junk result + matched affixes/aspects
alt kept equipment
    Vision->>Locator: locate_affix_markers(cropped tooltip, Item, matches)
    Locator-->>Vision: reliable marker centers
    Vision->>Overlay: draw keep outline + matched markers
else no match or ignored
    Vision->>Overlay: draw no-match/empty outline
end
```

</a>
</details>

<sub>Reviews (8): Last reviewed commit: ["Fix seals + charms"](https://github.com/d4lfteam/d4lf/commit/52dafb7babeb6b713c512c83c5a2e4c8bdc9832f) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=43546607)</sub>


src/item/descr/geometry_locator.py
```diff
@@ -0,0 +1,263 @@
...
+
+    # Seal bullets: only for HoradricSeal items or Set-rarity items.
+    if item.item_type == ItemType.HoradricSeal or item.rarity == ItemRarity.Set:
+        base.append("seal_set_bullet_point")
+
+    # Normal affix bullets are always needed.
+    base += [f"affix_bullet_point_{x}" for x in range(1, 3)]
+
+    # Rerolled bullets: only when TTS reports at least one rerolled affix.
+    if AffixType.rerolled in affix_types:
```
cjshrader: TTS has no way to determine if an affix is rerolled or tempered. This is actually backwards, we only know if an affix is rerolled or tempered based on the icon that matched it when we did the template search. That's actually a leftover from the full screen reading days. We will need to always include the rerolled and tempered bullets.

One could argue we don't need to keep the tempered bullets though because we don't match on tempered bullets. We don't even evaluate them. A user can't say "I want to keep all items with movement speed tempered on them".

src/item/descr/texture.py
cjshrader: What is the difference between your geometry locator and this method and the methods below it? Do they all need to still exist?

cjshrader: Well ok it uses this one but the two methods below this one.

cjshrader (changes requested):
Beyond my comments can you add an explanation of what exactly this does differently than the code we already have? I don't really get what it's doing. Does this remove the need for the search_threshold? Is it just narrowing the ROI? That wouldn't be enough to fix schoof's issues.

I'm thinking though that more code needs to be deleted and then I'll understand better what's going on. I think the method read_descr_mixed is no longer used anywhere, can you confirm? If it isn't then it and a lot of stuff that uses it can go away. There's also probably no reason to keep loc on the affix object anymore either. It looks like we don't calculate it ahead of time anymore

src/item/descr/geometry_locator.py
cjshrader: To be clear I say just keep the tempered bullets anyway and put them in the list of everything we search.

src/item/descr/texture.py
chrisoro: I'll answer it more broadly in a top level comment

chrisoro:
First of all the refactor hides all logic behind a deep interface which allows us to replace it and allowed me to try several solutions without changing 10000000 lines of code. We can now simply exchange the locator class and compare.

I kept the old stuff since I thought you still wanted it in for legacy reasons or to quickly change back but I'm also fine with a broader cleanup.

What the geometry locator does differently:

The old approach ran a second full item parse every time vision mode needed affix positions. It did a broad template search over all possible bullet types and then used which icon matched each row to determine the affix type. This needed up to 3 retries with 50 ms sleeps between them, and would fail with IndexError if the count was wrong.

The new AffixMarkerLocator reuses the Item already produced (no second parse!). Because TTS already told us the exact number of affixes, the item type, the rarity, and whether any greater affixes are present, it can build a minimal template list and validate success by expecting exactly len(inherent) + len(affixes) bullets. One pass, no retries, graceful failure instead of exceptions.
And since we no longer care which template matches, we can also reduce the threshold quite a bit even though some templates are very similar.

The speed gain (~70 ms → ~20 ms) comes from fewer templates searched and removing the retry loop.

src/item/descr/geometry_locator.py
```diff
@@ -0,0 +1,262 @@
...
+
+    base: list[str] = []
+
+    # Greater affix and masterworking bullets: only when TTS confirms greater affixes are present.
+    # TTS can identify greater affixes from their value format in the spoken text.
+    if AffixType.greater in affix_types:
+        base += [
+            "greater_affix_bullet_point_1",
+            "greater_affix_bullet_point_masterworked",
+            "masterworking_affix_bullet",
```
cjshrader: This is the cause of schoof's problem. There are two types of masterworked bullets. One for GAs, one for non GAs. masterworking_affix_bullet and the _2 of it are for non-GAs so they need to always be checked.

cjshrader (changes requested):

cjshrader:
To your overall comment, I'm curious if you're seeing very significant improvements now that you're searching for a lot more bullet points. The reparsing of the item was not a significant amount of time, only looking for bullets was a significant amount of time so removing the second reparse I don't think is helping too much.

Mainly my question is this, if we've lowered sensitivity, why don't we still have this problem where an 8 is matched?

image

Have you made the ROI smaller or something? It makes me worry that we're missing something.

src/scripts/vision_mode_with_highlighting.py
cjshrader: If you want to save a free 10 ms you could remove the gray outline. We really don't have a lot of "processing" time anymore and the queue is refreshed every 10 ms so you barely ever even see the gray outline.

cjshrader:
I see the answer to my top level comment. The build I gave schoof was not loading rerolled affix bullets. Unfortunately we have not solved the main problem with this. Either I can increase sensitivity and we miss some bullets, or I always could have lowered sensitivity and the 8 gets flagged. This is the same issue the old solution had, it's really not very different.

chrisoro:
Ok different approach: try all templates but clean findings based on their y position. stop early when unique findings are present. I'll leave the threshold with the updated low values but with this change we might be able to increase it back again

cjshrader:
This one still has issues with detecting the wrong bullets. The information for it is in discord though I can transfer that here if needed.

chrisoro:
Yea would be great if you can give me the matched templates and confidence on your test image

src/item/descr/geometry_locator.py
```diff
@@ -0,0 +1,172 @@
...
+            threshold=_BULLET_MATCH_THRESHOLD,
+            expected_count=len(item.inherent) + len(item.affixes),
+        )
+
+        if item.item_type == ItemType.HoradricSeal and affix_bullets:
+            affix_bullets = affix_bullets[1:]
+
+        expected_affix_rows = len(item.inherent) + len(item.affixes)
+        if len(affix_bullets) < expected_affix_rows:
+            return None
```
cjshrader, line 234: We really can't return none here. I know the error is annoying but this is the use case for the "not enough bullets" error and we need to be made aware of that instead of just hiding it. Plus it gives us the information we need to resolve it.

cjshrader:
Ok in this instance it is matching affix_bullet_point_1_medium. I am at 1920x1080 resolution and medium texture quality. score=0.9771950244903564 which of course makes sense since that is indeed an affix bullet, just in the wrong spot.

image

These issues of the bullet being in the wrong location are pretty consistent.

Ever since you updated the threshold to 0.8 I can not reproduce the issue of the 8 in the armor value getting marked. That's interesting because in the previous solution I had to update the threshold to 0.85 to prevent this so I'm not sure why 0.8 is solving it this time.

cjshrader:
As requested, here are the boots without the overlay. I doubt you'll be able to reproduce this with a static image, I'm sure the issue is grabbing icons before everything has loaded in. That would explain why it's inconsistent.

image

Here's another one, it very frequently highlights the second affix when it should be highlighting willpower.

image

This is happening on a lot of items so you should be able to log in and easily reproduce when you have the chance.

Honestly right now this solution is not really different than the previous solution except it doesn't work as well. It's really not doing much different at all. If you get tired of trying to make this work I am ok to go back to the previous solution and see what tweaks can be made there.

cjshrader:
I have not been able to cause any affix bullet issues at all at any resolution. The only problem I can cause is this one, which I can get to happen about 50% of the time with this one specific item in this one slot

image

Honestly it's entirely possible this would have happened before too. Looks like we're detecting the wrong location for the top left corner is all. I don't think it's anything we need to fix at this time.

I'm going to unleash this build on schoof and see how it goes. He has graphic issues I can never reproduce.

chrisoro:
Looks like our item border is matching the stash. the confidence is super low so give me the full screenshot without our filter

cjshrader:
image

Providing the screenshot as requested. I couldn't get it to happen at all again this morning for this specific item though I did get it 1 single time for a couple of other items but it's really not frequent so it's still fine with me if you just do nothing.

Here's another example of probably the same issue that I can get to happen a little bit more frequently

image

image

chrisoro:
@greptileai review

src/scripts/vision_mode_with_highlighting.py
```diff
@@ -388,39 +403,53 @@ def evaluate_item_and_queue_draw(self, item_descr: Item, cancel_event: Event) ->
...
+                                locator_result = None
                                 if any(
                                     res_matched.profile.endswith(ASPECT_UPGRADES_LABEL) for res_matched in res.matched
                                 ):
                                     self.request_codex_upgrade_box(item_descr, item_roi, res)
                                 else:
-                                    self.request_match_box(item_descr, item_roi, res, item_descr_with_loc)
+                                    if not is_sigil(item_descr.item_type):
+                                        matched_affixes = res.matched[0].matched_affixes if res.matched else []
+                                        aspect_matched = any(m.aspect_match for m in res.matched)
```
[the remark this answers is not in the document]
chrisoro: @greptileai this is by design. When multiple filters match, we just want to draw boxes based on the first filter

cjshrader:
You are now reading tempers, but I don't think this is correct or wanted behavior. We never parsed these before, and anyone can put any temper they want on an item so there's no reason to match on it. For example, this item:

image
image

It's matching on max life:

SEED OF HORAZON -- Matched maxroll s14 warlock minion warlock endgame midgame.Affixes.Amulet: ['willpower', 'critical_strike_chance', 'maximum_life', 'to_demonology_skills']

But it only has max life because I tempered it on. If I rerolled it on then sure, makes sense that I'd want to match, but if I temper it on there's no randomness there. 

I just see no positive benefit to that change. It's a change to fundamentally how we parse items. Maybe one day if we add temper support as something separate it would make sense, but not right now.

cjshrader:
These boots matched but no affixes were highlighted

image

image


TTS Found: ['SADISTIC WARLORD BOOTS', 'Legendary Boots', '850 Item Power', 'Armory Loadout', '638 Armor', '+91 Willpower +[83 - 99]', '+1,033 Maximum Life [1,016 - 1,225]', '+914 Armor [780 - 980]', '+24% Movement Speed [20 - 24]%', '+22% Movement Speed [15 - 25]%', 'Imprinted: When killing your Demons, you have a 45% [30 - 50]% chance for each to count as two kills.', 'Requires Level 70. Account Bound. Lord of Hatred Item', 'Sell Value: 18,733 Gold', 'Durability: 50/100. Tempers: 2/3', 'Right mouse button']
Parsed item based on TTS: Item(affixes=[Affix(loc=None, max_value=99.0, min_value=83.0, name='willpower', text='+91 Willpower +[83 - 99]', type=<AffixType.normal: 3>, value=91.0), Affix(loc=None, max_value=1225.0, min_value=1016.0, name='maximum_life', text='+1,033 Maximum Life [1,016 - 1,225]', type=<AffixType.normal: 3>, value=1033.0), Affix(loc=None, max_value=980.0, min_value=780.0, name='armor', text='+914 Armor [780 - 980]', type=<AffixType.normal: 3>, value=914.0), Affix(loc=None, max_value=24.0, min_value=20.0, name='movement_speed', text='+24% Movement Speed [20 - 24]%', type=<AffixType.normal: 3>, value=24.0), Affix(loc=None, max_value=25.0, min_value=15.0, name='movement_speed', text='+22% Movement Speed [15 - 25]%', type=<AffixType.normal: 3>, value=22.0)], aspect=Aspect(name='sadistic', loc=None, min_value=None, max_value=None, text='Imprinted: When killing your Demons, you have a 45% [30 - 50]% chance for each to count as two kills.', value=None), codex_upgrade=False, cosmetic_upgrade=False, inherent=[], is_ancestral=False, is_in_shop=False, item_type=<ItemType.Boots: 'boots'>, name='sadistic_warlord_boots', original_name='SADISTIC WARLORD BOOTS', power=850, rarity=<ItemRarity.Legendary: 'legendary'>, seasonal_attribute=None, set=None)
SADISTIC WARLORD BOOTS -- Matched maxroll s14 warlock minion warlock endgame midgame.Affixes.Boots: ['movement_speed', 'willpower', 'maximum_life', 'armor']

There is no error in the log. 

Here's the profile if that's useful:

```yaml
- Boots:
    affix_pool:
    - count:
      - {name: movement_speed}
      - {name: willpower}
      - {name: maximum_life}
      - {name: armor}
      - {name: to_rampage}
      - {name: wrath_regeneration}
      min_count: 3
    item_type: [boots]
    min_power: 850
```

Edit: now the boots are working fine? Strange

chrisoro:
@greptileai review

cjshrader (approved):
I got the impression this is all done with testing and it's definitely been tested enough

PR #832, opened 2026-07-10
(BugFix) Handle invalid window process ids & reset window state when D4 closes. Should resolve pid must be a positive integer error.

### LINKED ISSUE #801: ValueError: pid must be a positive integer (fatal crash)
Now I am running `d4lf.exe` as it is intended (instead of my typical console log).
And I deleted my previoys params.ini and let it be created by the system and never touched it by hand.

This morning I noticed stange crash after I changed some d4lf settings (line 33 2026-07-03 08:37) and then later started playing (line 34 2026-07-03 09:05).
AFAIK I started Diablo 4 after I had changed the settings, so D4 was not running when I started `d4lf.exe` (line 1 2026-07-03 08:37).

This is just for yout information and first time this happened to me.
It might be some changes you have made recently that caused this or something else...
The call stack is quite explicit about was happened so it might be easy to fix this by retrying after you call `_get_process_from_window_name` again???

Image

[log_2026_07_03_05_37_34.log](https://github.com/user-attachments/files/29624550/log_2026_07_03_05_37_34.log)

chrisoro:
fixes https://github.com/d4lfteam/d4lf/issues/801


cjshrader:
@chrisoro can you please merge main into this. I tried to do it and botched it to be honest. I believe I have reverted my changes.

cjshrader (approved):
This one is tough because the root cause has been unclear, so I never reproduced it reliably so I don't know if this fixed it. But it doesn't appear to have broken anything.

chrisoro:
Well the tests to reproduce the issue. He might have more but this was definitely one of them

PR #837, opened 2026-07-19
Refactor structure

chrisoro:


cjshrader:
I obviously find this very difficult to read, but that's of course because I've worked on one code base for years and now this is a different code base. It'll take getting used to. Some things that popped out to me.

Under desktop, there are 3 files just named "core.py". They aren't very clear what they do at all.

I'm not clear if this is an artifact of how I checked it out but it looks like folders like "config" and "gui" need to be outright deleted, their contents are empty but I still see an empty folder.

The loot/colors.py has just one item that does colors. But it mostly has stuff like "mark as junk" etc, that name is very confusing.

Same with things like loot/fast and loot/highlighting, if they were in something like vision/fast and vision/highlighting it'd make more sense but they're just in the middle of a bunch of files. I'd think the ideal solution would be to make an "overlay" package and put the main entry to each overlay in there (So paragon and the info as well). Though I see there is an overlay package.

Perception is a weird name for a package, I guess that's all the screen reading stuff? I'd probably just call it screenreading or something.

One thing I'd like to see in a refactor is we have colors all over the code base. It'd be nice if they were centralized somewhere.

I wasn't a fan of what we had before but honestly the more I look at this the more I think this might be worse. It feels all over the place. Under profiles is both the sigils model, and also the sigils GUI elements, for example. That doesn't seem right.

If this is what's needed for AI review, I might be about done with d4lf entirely honestly. This isn't a weird ultimatum, you understand LLMs for this kind of thing better than I do, but I just think this won't be very workable for a human anymore. If that's what you think will work best then we should go with that.

PR #849, opened 2026-09-18
[FIX] - startup crash and non-working hotkeys on Python 3.14

### LINKED ISSUE #848: D4LF Crashing on latest version
Image

Maybe is related to the new season changes, or maybe is something within my own configuration.

I generated the profiles for the new season for the barbarian and it gave some error while loading, can't remember what, but seems something related to "transfigure" and the new "Insight" unique.

After closing it (TTS wasn't working) and reopening it, this crash started happening on the console only mode, and vision mode won't work (also all hotkeys won't work) on UI mode:

Image

guillepinho:
Fixes #848.

## What happens

D4LF crashes on startup in console-only mode. In UI mode the process survives, but no hotkey works and vision mode can't be started.

File "src\settings\hotkeys\runtime.py", line 63, in annotate NameError: name 'Hashable' is not defined. Did you mean: 'callable'?

The above exception was the direct cause of the following exception:

File "src\app\handler.py", line 212, in setup_key_binds
File "src\settings\hotkeys\runtime.py", line 96, in _restart_listener
File "pynput\keyboard\_win32.py", line 292, in init
File "pynput\_util\init.py", line 291, in _wrap
File "inspect.py", line 1282, in getfullargspec
TypeError: unsupported callable

## Why

`src/settings/hotkeys/runtime.py` imported `Callable` and `Hashable` inside an `if TYPE_CHECKING:` block, but `Hashable` is used in the annotations of `_on_press` and `_on_release`. Under PEP 649 (Python 3.14, which this project requires) those annotations aren't evaluated at definition time, they become a lazy `__annotate__`. Nothing notices until someone introspects the function.

pynput does exactly that. `_wrap` in `pynput/_util/__init__.py:291` calls `inspect.getfullargspec()` on the handler to decide how many arguments to pass it. That forces annotation evaluation, `Hashable` doesn't exist at runtime, and `getfullargspec` turns the resulting `NameError` into `TypeError: unsupported callable`.

One cause, two symptoms. In console-only mode the exception propagates to `main.py` and kills the process. In UI mode `setup_key_binds` dies on its first `_register_hotkey` call, so no global listener is ever created and every hotkey
is dead, including the vision mode one. TTS still connects because it's a separate path, which matches the "TTS: Connected / Vision Mode: STOPPED" state in the report.

This is a regression from #844, not a season 15 issue.

## Why CI didn't catch it

`tests/settings/hotkeys/runtime_test.py` patches `keyboard.Listener` with `_FakeListener`, which stores the callbacks without inspecting them. pynput's `_wrap` never runs under test, so all 14 existing tests pass against the broken code.

## The fix

Move both imports to module scope. The ruff rule that flags this (`typing-only-standard-library-import`) is what pushes the imports into the `TYPE_CHECKING` block in the first place, so it needs an explicit suppression with the reason attached, otherwise the next person to run `ruff --fix` reintroduces the crash.

Added a regression test that calls `inspect.getfullargspec` on both handlers, hitting the real mechanism instead of the mock. It fails on `main` with `TypeError: unsupported callable` and passes with the fix.

## Verification

Reproduced against the released v10.0.3 `d4lf.exe` on Windows, same traceback as the report. Then ran a minimal repro on Windows with Python 3.14.7 and the real win32 backend, constructing an actual `keyboard.Listener`:

| | `main` | this PR |
|---|---|---|
| `getfullargspec(_on_press)` | `TypeError: unsupported callable` | ok |
| `add_hotkey` + win32 `Listener` | `TypeError: unsupported callable` | ok |

Full suite: 1126 passed, 26 skipped. `ruff check`, `ruff format` and `ty check` all clean.

<h2><a href="https://app.greptile.com/api/retrigger?id=66031778"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/RetriggerDark.svg?v=2"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/Retrigger.svg?v=2">Retrigger</picture></a>Confidence Score: 5/5</h2>

Safe to merge; the changed listener-handler initialization path completed successfully.

What we checked:
- Ran the contract validation for the pull request and confirmed the _on_press and _on_release handlers return FullArgSpec with args ['self', 'key'] and annotations including return: None and key as collections.abc.Hashable. <a href="https://www.greptile.com/trex">T-Rex</a>
- Observed that the listener was constructed as pynput.keyboard._base.Listener and that the test reported PASS with both bound handlers inspectable and accepted by keyboard.Listener (exit code 0). <a href="https://www.greptile.com/trex">T-Rex</a>
- Verified that the focused change moves Callable and Hashable out of TYPE_CHECKING, enabling annotations to be resolvable by inspect on Python 3.14. <a href="https://www.greptile.com/trex">T-Rex</a>
- Collected and attached artifacts documenting the results, including the Python signature artifact and the two validation logs. <a href="https://www.greptile.com/trex">T-Rex</a>

<details open><summary>Summary</summary>

This update makes hotkey callback annotation types available when the keyboard library inspects listener handlers on Python 3.14. The focused runtime check confirmed both handlers are inspectable and accepted when creating a listener. Safe to merge.
</details>

<sub>Reviews (1) · Last reviewed commit: ["refactor: streamline imports and enhance..."](https://github.com/d4lfteam/d4lf/commit/678af61af558ec62d56c58dc307e26c25e5375a8)</sub>


chrisoro:
Thank you!

chrisoro (approved):

